Issue #28, opened 2026-07-30
docs: add docs/CLI.md — moon_cli.py has no documentation page


LeyckerS:
### Context

The Documentation table in the README lists eight pages — Quick start, Configuration, Providers, Architecture, Engineering notes, Troubleshooting, FAQ, Changelog. None of them cover `moon_cli.py`.

That is a gap, because the Architecture section presents the CLI as a first-class front-end sitting at the same level as the GUI:

```
moon_cli.py       headless CLI, same engine, same extraction layer
```

Someone who wants to script MoonDownloader (cron, batch job, CI) currently has to read the source to find out which flags exist.

### What to do

Create `docs/CLI.md` containing:

1. **Every flag** `moon_cli.py` accepts — name, type, default, what it changes. Read them off the argument parser in `moon_cli.py`; do not guess.
2. **Exit codes** — document what the CLI returns today. Describe current behaviour only; changing the exit codes is out of scope here.
3. **Three worked examples:**
   - a batch of `fuckingfast.co` links (no browser is launched)
   - a batch containing at least one `datanodes.to` link (one shared Chrome, Turnstile wait)
   - a mixed batch
4. **Relationship to the GUI settings** — which CLI flag maps to which panel setting, linking to `docs/CONFIGURATION.md` instead of duplicating it.

Then add one row to the Documentation table in `README.md`:

```markdown
| [CLI](docs/CLI.md) | every flag, exit codes, scripting examples |
```

### Acceptance criteria

- [ ] Every flag present in `moon_cli.py` appears in `docs/CLI.md`
- [ ] No flag documented that does not exist in the source
- [ ] `README.md` Documentation table links the new page
- [ ] Style matches the existing pages in `docs/`

### Notes for the contributor

**You do not need Windows or a working download to close this** — it is a read-the-source-and-write-it-down task. The CLI argument parser is the only thing you have to read.

Scope is roughly 80–150 lines of Markdown. See [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md) for style conventions. Comment here before you start so two people don't write the same page.

kushin25:
I will prepare the requested CLI reference from the current argument parser, including only documented current behavior. This contribution is being prepared with automated assistance; I will run the repository’s applicable local validation before opening a PR.

LeyckerS:
Done in #37 by @kushin25. `docs/CLI.md` is on `main` and the Documentation table links it.

[...]

Issue #32, opened 2026-07-31
enhancement: give moon_cli.py meaningful exit codes for scripting


LeyckerS:
### Context

`moon_cli.py` exists so MoonDownloader can run unattended — cron, a scheduled task, a step in a larger script. That only works if the caller can tell success from failure without parsing the log.

Today the CLI does not distinguish outcomes in its exit status, so a run where 3 of 29 files failed looks identical to a clean run from the outside.

### What to do

Define and implement three exit codes in `moon_cli.py`:

| Code | Meaning |
|:--|:--|
| `0` | every file in the batch completed |
| `1` | the run completed but at least one file failed (dead link, exhausted retries, extraction failure) |
| `2` | the run could not start — bad arguments, unreadable link file, invalid configuration |

Rules:

- **A partial failure is still exit 1, not an exception.** The remaining files must still be attempted; the code reports the outcome, it does not abort the batch.
- **Exit 2 is pre-flight only.** Once the first download starts, the outcome is 0 or 1.
- Keep the existing summary line on stdout unchanged — scripts may already be grepping it.

### Acceptance criteria

- [ ] A batch where every file succeeds exits 0
- [ ] A batch containing one guaranteed-dead link exits 1, and the surviving files still download
- [ ] An unreadable link file exits 2 before any network activity
- [ ] A test in `tests/` covers all three, with the network stubbed the way `test_no_chrome.py` already does
- [ ] The table above is documented in `docs/CLI.md`

### Depends on

#28 creates `docs/CLI.md`. If that page does not exist yet when you open the PR, add the exit-code table to your PR and note it — the two will be reconciled on merge.

### Notes for the contributor

This one touches real behaviour rather than docs or config, so read `moon_cli.py` end to end first. `CONTRIBUTING.md` has the rules that are not style preferences — in particular, `moon_cli.py` currently carries its own copy of the download engine, so check whether your change needs to land in `moon_engine.py` too.

Comment here before you start.

LeyckerS:
Note for whoever takes this — the current behaviour is narrower than "always returns 0", and knowing where the gap actually is will save you reading the whole file.

`moon_cli.py` **does** exit non-zero in three places, all of them before or around the run rather than about its outcome:

- `moon_cli.py:319` — URL file not found
- `moon_cli.py:325` — file parsed but no URLs in it
- `moon_cli.py:342` — an unexpected top-level exception, after writing `crash_log.txt`

[...]

PR #37, opened 2026-07-31
docs: add CLI reference

kushin25:
## Description

Adds the CLI reference requested in #28. It documents every current `moon_cli.py` argument, current exit-code behavior, three scripting examples, and the relationship between CLI flags and GUI settings. README now links the page.

Related issue: #28

This contribution was prepared with automated assistance; the documented behavior was checked against the current argument parser and source.

## Type of change

- [x] Documentation update

## Checklist

- [x] This does not affect shared logic (extraction, download engine)
- [x] I have not added new dependencies

## Validation

- `git diff --check`
- `.venv/bin/python moon_cli.py --help`
- `.venv/bin/python test_no_chrome.py`
- `python3 -m py_compile moon_cli.py`

The no-Chrome test passed: fuckingfast-only batches used zero browsers and a mixed batch started one shared Chrome instance.

## Notes

No live downloads were run; example URLs are illustrative and the change is documentation-only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,7 @@ Full write-up: **[docs/ARCHITECTURE.md](docs/ARCHITECTURE.md)**.
 | Document | What's in it |
 |:--|:--|
 | [Quick start](docs/QUICKSTART.md) | install, first run, what each setting does |
+| [CLI](docs/CLI.md) | every flag, exit codes, scripting examples |
 | [Configuration](docs/CONFIGURATION.md) | every setting, every environment variable, the Chrome profile |
 | [Providers](docs/PROVIDERS.md) | how each host is extracted, and how to add another |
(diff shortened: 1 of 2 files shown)


LeyckerS:
Merging this. Every flag, default and exit code in here matches the source — I checked each one.

The detail that stood out: you documented that `moontech_cli_*.log` and `.json` land beside `moon_cli.py` rather than inside `--output`. The parameter in `Telemetry.save()` is literally named `out_dir`, so the obvious reading is the output folder — you only get the right answer by following the call site to line 563 where `base` is the script directory. Nice.

Catching that `Ctrl+C` exits 0 rather than non-zero was the other one most people would have missed.

If you want another: #32 changes those exit codes to something scriptable, and now that this page exists it's the natural follow-up. Yours if you want it.

Issue #34, opened 2026-07-31
refactor: move the duplicated download engine into a shared module


LeyckerS:
### Context

This is documented debt — `CONTRIBUTING.md` states it outright:

> **`moon_engine.py` and `moon_cli.py` still carry their own copy of the download engine** (`download_file`, `Telemetry`, `ProxyPool`). A fix to one belongs in both until that code moves into a shared module.

That rule exists because the duplication is real, and it makes every bug fix in the download path a two-file change that a contributor has to know about in advance. Miss the second copy and the GUI and the CLI silently diverge — the kind of bug that only shows up for whichever front-end you were not testing.

The extraction layer already solved this problem correctly: `moon_extract.py` is shared, and `BrowserGate` lives there precisely so the launch decision cannot differ between front-ends. The download engine never got the same treatment.

### What to do

Extract the duplicated pieces into a single module — `moon_download.py` alongside `moon_extract.py`:

- `download_file` — including Range-header resume and stall detection
- `Telemetry`
- `ProxyPool`

Then have both `moon_engine.py` and `moon_cli.py` import from it.

**Constraints:**

- **Behaviour must not change.** This is a move, not a rewrite. If the two copies have drifted, the diff between them goes in the PR description and the resolution gets discussed before it lands — do not silently pick one.
- **`FileRecord` publishing stays as it is.** `download_file` publishes `rec.done_bytes` and `rec.live_mbs` on its own ~4 Hz window, deliberately separate from the stall detector's 60 s history. Keep those two clocks independent.
- **Update `CONTRIBUTING.md`** — the rule quoted above is what this issue removes.

### Acceptance criteria

- [ ] `download_file`, `Telemetry` and `ProxyPool` exist in exactly one place
- [ ] `moon_engine.py` and `moon_cli.py` both import them; neither defines them
- [ ] Any drift between the two former copies is documented in the PR description
- [ ] CI green, `test_no_chrome.py` still passes
- [ ] A GUI run and a CLI run over the same batch produce the same result
- [ ] The obsolete rule is removed from `CONTRIBUTING.md`

### Notes for the contributor

**This is the largest open item in the repo and it is not a first issue.** It needs someone willing to read both files properly and diff them line by line. Verification realistically wants Windows and live links per `CONTRIBUTING.md` — say so when you claim it, and we can split the review.

Comment here before you start so this does not get duplicated.

LeyckerS:
[...]

PR #38, opened 2026-07-31
test: move no-chrome regression to pytest

### LINKED ISSUE #33: test: move test_no_chrome.py into a pytest suite under tests/
### Context

CI runs `test_no_chrome.py` as a plain script:

```yaml
- name: A fuckingfast batch must not open a browser
  run: python test_no_chrome.py
```

It works, but it is one script with one exit code. There is no per-case reporting, no `-k` to run a single case, and when it fails the log does not say which assertion broke without reading the source.

`CONTRIBUTING.md` also lists three more verification scripts that nothing runs automatically:

```bash
python test_no_chrome.py        # no browser for fuckingfast, exactly one for datanodes
python integration_http.py      # browser -> loopback HTTP -> engine
python integration_web.py       # pywebview path
python render_gui.py out/       # GUI renders + overflow audit
```

### What to do

**1. Move and split the existing test**

- `test_no_chrome.py` → `tests/test_no_chrome.py`
- Split the script body into `test_*` functions, one per assertion it already makes
- Move the Chrome and network stubs into `tests/conftest.py` as fixtures — they are already written, they just need to become fixtures

**2. Point CI at pytest**

```yaml
- run: pip install "pytest==8.*"
- run: pytest tests/ -q
```

**3. Do not change what is being asserted.** The two guarantees stay exactly as they are: a `fuckingfast.co` batch launches no browser, a batch with a `datanodes.to` link launches exactly one. If a rewrite makes an assertion weaker, it is wrong.

### Out of scope

`integration_http.py`, `integration_web.py` and `render_gui.py` stay where they are. They need a display or a real browser and do not belong in this change — a follow-up can look at which of them can run headless.

### Acceptance criteria

- [ ] `pytest tests/ -q` green locally and in CI
- [ ] Every assertion present in the old script is still made
- [ ] The `no-chrome` CI job runs pytest instead of the bare script
- [ ] `CONTRIBUTING.md` verification block updated to the new command
- [ ] No new dependency beyond `pytest`

### Notes for the contributor

**No Windows needed** — the whole point of this test is that Chrome and the network are stubbed at the `moon_extract` boundary, so it runs anywhere. Comment here before you start.

pollychen-lab:
## Summary

Moves the no-Chrome regression coverage from the standalone `test_no_chrome.py` script into a pytest suite under `tests/`. The Chrome/network stubs are now pytest fixtures, while the existing guarantees remain split into per-case assertions for the engine, CLI, and source-level BrowserGate routing.

Updates CI to install pytest and run `pytest tests/ -q`, and updates the contributing verification command accordingly.

Fixes #33.

## Testing

- `python3 -m pytest tests/ -q`
- `python3 -m compileall -q moon_engine.py moon_cli.py moon_extract.py moon_bridge.py render_gui.py integration_http.py integration_web.py tests/conftest.py tests/test_no_chrome.py`
- `git diff --check`


LeyckerS:
CI green on both jobs — merging.

What sold this: you split the three original guarantees into five per-case tests instead of collapsing them, so the engine path and the CLI path are now asserted separately rather than in one combined run. The static `BrowserGate` check survived the move intact too, which is the assertion that actually stops the old regression from coming back.

Also appreciated the small things: `pytest` pinned rather than floating, the `compileall` list in `lint.yml` updated to match the new layout instead of left stale, and the verification block in `CONTRIBUTING.md` fixed in the same PR so the docs don't drift from the command.

If you want another, #34 is the real one — `moon_engine.py` and `moon_cli.py` each carry their own copy of the download engine, and now that there's a test suite to move under, it's finally approachable.

Issue #26, opened 2026-07-30
ci: run the matrix on Python 3.11 and 3.12, add a ruff check


LeyckerS:
### Context

The README states the requirement as **Python 3.10+**, but CI only ever runs 3.10:

```yaml
# .github/workflows/lint.yml — both jobs
- name: Set up Python
  uses: actions/setup-python@v6
  with:
    python-version: "3.10"
```

Nothing verifies that the engine imports cleanly on 3.11 or 3.12. Anyone on a current Python install is running an untested configuration.

Second gap: the `syntax-check` job only runs `python -m compileall -q`, which catches syntax errors and nothing else. Unused imports, undefined names, shadowed builtins — all invisible today.

### What to do

**1. Matrix the `syntax-check` job**

```yaml
jobs:
  syntax-check:
    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
      matrix:
        python-version: ["3.10", "3.11", "3.12"]
    steps:
      - uses: actions/checkout@v7
      - uses: actions/setup-python@v6
        with:
          python-version: ${{ matrix.python-version }}
      # ... existing compileall step unchanged
```

`fail-fast: false` matters — if 3.12 breaks we still want to see whether 3.11 passed.

**2. Add a ruff step**

Add `ruff.toml` at the repo root and a `ruff check .` step. Start from a conservative rule set that passes on the code as it stands today:

```toml
line-length = 100
target-version = "py310"

[lint]
select = ["E", "F", "W", "I"]
ignore = ["E501"]
```

**Do not reformat the codebase in this PR.** If a rule produces more than a handful of hits, add it to `ignore` and open a separate issue for it. The goal is a green baseline that catches regressions, not a 1300-line diff.

**3. Leave the `no-chrome` job alone.** It stays on a single Python version — it is a behavioural smoke test, not a compatibility check.

### Acceptance criteria

- [ ] Three matrix jobs, all green
- [ ] `ruff check .` green with zero source changes outside `ruff.toml`
- [ ] `no-chrome` job still passes
- [ ] `ruff` pinned to a specific version in the workflow step

### Notes for the contributor

**You do not need Windows for this** — everything runs on `ubuntu-latest`. You can validate the whole thing from a fork before opening the PR.

Scope is roughly 25 lines of YAML plus a small TOML file. Comment here before you start.

Moferanoluwa:
[...]

PR #41, opened 2026-07-31
refactor: move download engine into shared module

### LINKED ISSUE #34: refactor: move the duplicated download engine into a shared module

pollychen-lab:
## Description

Extracts the duplicated download engine from `moon_engine.py` and `moon_cli.py` into a new single-file module, `moon_download.py`. Both frontends now import the shared `download_file`, `Telemetry`, `FileRecord`, `ProxyPool`, proxy/session helpers, and download tuning constants from that module.

Fixes #34.

Drift found while moving the code:

- The GUI engine copy had live `FileRecord.done_bytes` / `FileRecord.live_mbs` publishing for row progress; the CLI copy did not expose those fields. The shared implementation keeps the live publishing so the GUI behavior stays intact, while the CLI report writer preserves the old shorter JSON field set.
- The GUI `Telemetry` copy recorded browser counts, stall events, slow-file summaries, and detailed concurrency logs; the CLI copy wrote `moontech_cli_*` files with a shorter summary. The shared `Telemetry` keeps both formats via an engine/default flavor and a CLI flavor.
- The GUI `download_file` accepted `Telemetry` so stall kills could be recorded; the CLI path did not. The shared function keeps telemetry optional and the GUI passes it explicitly.

`CONTRIBUTING.md` is updated to remove the obsolete two-copy rule and document `moon_download.py` as the shared download/telemetry layer.

## Type of change

- [x] Documentation update
- [x] Refactor / code cleanup
- [ ] Other:

## Checklist

      description

Local validation:

- `python3 -m compileall -q moon_engine.py moon_cli.py moon_download.py moon_extract.py moon_bridge.py render_gui.py integration_http.py integration_web.py tests/conftest.py tests/test_no_chrome.py`
- `python3 -m pytest tests/ -q`
- `git diff --check`

The pytest suite covers the stubbed GUI engine and CLI no-browser paths and now asserts that `download_file`, `Telemetry`, and `ProxyPool` are defined only in `moon_download.py`.

I did not run the live Windows/browser verification with real provider links in this environment.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,6 +36,7 @@ moon_bridge.py     loopback HTTP + token, launches Edge/Chrome --app, OS dialogs
   moon_engine.py   the engine with no GUI: start/stop/snapshot
     moon_extract.py  datanodes (real Chrome over CDP) · fuckingfast (curl_cffi)
                      BrowserGate: the launch, deferred until a datanodes link
+    moon_download.py download_file · Telemetry · ProxyPool
 
 moon_cli.py        argparse CLI, same engine, same extraction layer
 ```
@@ -43,21 +44,19 @@ moon_cli.py        argparse CLI, same engine, same extraction layer
 Layers inside the engine:
 
 - **Extraction** — `moon_extract.py`, shared by all three front-ends
-- **Download engine** — aiohttp, Range-header resume, stall detection, proxy rotation
-- **Telemetry** — 1 Hz snapshots, `.txt` + `.json` output
+- **Download engine** — `moon_download.py`, shared by the GUI engine and CLI
+- **Telemetry** — `moon_download.py`, 1 Hz snapshots, `.txt` + `.json` output
 - **GUI** — `web/` over the loopback API, hosted by `moon_bridge.py`
 
 ## Rules that are not style preferences
 
-- **Shared logic goes in `moon_extract.py`**, not copy-pasted between front-ends. If a
-  change touches extraction or the Chrome lifecycle, it must land in one place and be
-  visible from both `moon_engine.py` and `moon_cli.py`.
+- **Shared logic goes in `moon_extract.py` or `moon_download.py`**, not copy-pasted between
+  front-ends. If a change touches extraction, the Chrome lifecycle, downloading,
+  telemetry or proxy rotation, it must land in one place and be visible from both
+  `moon_engine.py` and `moon_cli.py`.
 - **Never open a browser before you know you need one.** Ask `BrowserGate.get()` inside
   the provider branch that requires it. A launch at the top of a run is the bug
   `test_no_chrome.py` exists to prevent.
-- **`moon_engine.py` and `moon_cli.py` still carry their own copy of the download engine**
-  (`download_file`, `Telemetry`, `ProxyPool`). A fix to one belongs in both until that code
-  moves into a shared module.
 - **No new dependencies without a strong reason.** The stack is deliberately small:
   `aiohttp`, `playwright`, `curl_cffi`.
 - **English only.** Code, comments, log lines, dialog titles and docs. The GUI's EN/IT
@@ -72,7 +71,7 @@ python integration_web.py      # pywebview path
 python render_gui.py out/           # GUI renders + overflow audit
 ```
 
-`test_no_chrome.py` stubs Chrome and the network at the `moon_extract` boundary, so it
+`tests/test_no_chrome.py` stubs Chrome and the network at the `moon_extract` boundary, so it
 needs no browser, no display and no Playwright install.
 
 Live testing: at least 10 links per provider, including one guaranteed-dead one so
--- a/moon_cli.py
+++ b/moon_cli.py
@@ -7,201 +7,18 @@ Usage:
     python moon_cli.py --urls links.txt --output /path/to/downloads
     python moon_cli.py --urls links.txt --output ./dl --browsers 8 --streams 24 --retries 3
 """
-import os, re, sys, asyncio, threading, argparse, json, datetime
-import time, random, traceback, collections, io
+import os, sys, asyncio, threading, argparse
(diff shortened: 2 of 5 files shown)


LeyckerS:
This is a serious piece of work and the review reflects that — it is going to take me a bit longer than the others, so here is where it stands rather than leaving you waiting.

**What is right:**

- The three divergences between the two copies are documented in the description instead of silently resolved. That was the actual hard requirement in #34, and it is the part I expected to have to ask for.
- `FileRecord.done_bytes` / `live_mbs` live publishing preserved for the GUI, with the CLI keeping its shorter JSON field set — correct call on both sides.
- `test_front_ends_import_shared_download_engine` is the right instinct: it makes the deduplication permanent rather than something that quietly re-forks in six months. That test is arguably worth more than the refactor itself.
- `CONTRIBUTING.md` rule removed in the same PR, so the docs do not outlive the debt.
- Net −167 lines, CI green on both jobs.
- And you said plainly that you did not run the live Windows/browser verification. Saying so is worth more than a checkbox ticked without meaning it.

**Why I am not merging it tonight:**

This is the download path. CI here proves imports resolve and the `BrowserGate` behaviour holds with Chrome and the network stubbed — it does not prove a single byte still lands on disk. `CONTRIBUTING.md` asks for at least 10 links per provider including a guaranteed-dead one, plus a 40+ file session to hit the concurrency paths, and that has to happen on Windows against live providers. I need to run that myself before this goes to `main`.

Nothing for you to do about that — it is my job, not yours.

**One thing that may come back to you:** #48 changes `ProxyPool` in both front-ends. If I take that one first, this will need a rebase with the proxy change carried into `moon_download.py`. I will tell you either way rather than leaving it to a surprise conflict.

Trailing whitespace on 5 added lines — trivial, mentioning it only because #26 will add `ruff` to CI.

pollychen-lab:
> This is a serious piece of work and the review reflects that — it is going to take me a bit longer than the others, so here is where it stands rather than leaving you waiting.
> 
> **What is right:**
> 
>     * The three divergences between the two copies are documented in the description instead of silently resolved. That was the actual hard requirement in [refactor: move the duplicated download engine into a shared module #34](https://github.com/LeyckerS/moondownloader/issues/34), and it is the part I expected to have to ask for.
> 
>     * `FileRecord.done_bytes` / `live_mbs` live publishing preserved for the GUI, with the CLI keeping its shorter JSON field set — correct call on both sides.
> 
>     * `test_front_ends_import_shared_download_engine` is the right instinct: it makes the deduplication permanent rather than something that quietly re-forks in six months. That test is arguably worth more than the refactor itself.
> 
>     * `CONTRIBUTING.md` rule removed in the same PR, so the docs do not outlive the debt.
> 
>     * Net −167 lines, CI green on both jobs.
> 
>     * And you said plainly that you did not run the live Windows/browser verification. Saying so is worth more than a checkbox ticked without meaning it.
> 
> 
> **Why I am not merging it tonight:**
> 
> This is the download path. CI here proves imports resolve and the `BrowserGate` behaviour holds with Chrome and the network stubbed — it does not prove a single byte still lands on disk. `CONTRIBUTING.md` asks for at least 10 links per provider including a guaranteed-dead one, plus a 40+ file session to hit the concurrency paths, and that has to happen on Windows against live providers. I need to run that myself before this goes to `main`.
> 
> Nothing for you to do about that — it is my job, not yours.
> 
> **One thing that may come back to you:** #48 changes `ProxyPool` in both front-ends. If I take that one first, this will need a rebase with the proxy change carried into `moon_download.py`. I will tell you either way rather than leaving it to a surprise conflict.
> 
> Trailing whitespace on 5 added lines — trivial, mentioning it only because #26 will add `ruff` to CI.

@LeyckerS understand and thank you!

Issue #46, opened 2026-07-31
cleanup: 30 `except Exception:` handlers have no comment, against the documented rule


LeyckerS:
### Context

`CONTRIBUTING.md` states the rule:

> No blanket `except:` — name the exception, or `except Exception:` with a comment when the swallow is deliberate.

The rule is right. **30 handlers in the codebase do not follow it**, swallowing everything with no comment saying why.

These numbers were recounted against `main` on 1 August 2026, after #54 annotated its five. An earlier version of this issue said 25, which was wrong in both directions — `moon_extract.py` alone has 23, and three of the files #54 touched still have unannotated handlers elsewhere.

| File | Count | Lines |
|:--|--:|:--|
| `moon_extract.py` | 23 | 101, 188, 378, 391, 471, 537, 556, 653, 713, 735, 781, 796, 802, 806, 943, 1025, 1077, 1088, 1095, 1098, 1179, 1219, 1249 |
| `moon_engine.py` | 3 | 265, 371, 505 |
| `moon_bridge.py` | 2 | 170, 297 |
| `moon_cli.py` | 1 | 207 |
| `moon_download.py` | 1 | 564 |

Most are almost certainly deliberate — this is a codebase that talks to a browser over CDP and to hosts that change behaviour without notice, so defensive swallowing is often the correct call. That is exactly why the rule asks for a comment: **from the outside, a deliberate swallow and a forgotten one look identical**, and the next person cannot tell which is which.

The `moon_extract.py` ones matter most. Several of them currently mean a broken selector is indistinguishable from a dead link, which is the class of bug that costs someone an afternoon of debugging the wrong thing.

### Take a slice — four people can work in parallel

Claim **one slice** by commenting which you are taking. Do not take two.

| Slice | Scope | Handlers |
|:--|:--|--:|
| **A** | `moon_extract.py` lines 101–556 | 7 |
| **B** | `moon_extract.py` lines 653–943 | 8 |
| **C** | `moon_extract.py` lines 1025–1249 | 8 |
| **D** | `moon_engine.py`, `moon_bridge.py`, `moon_cli.py`, `moon_download.py` | 7 |

Slice D touches four files but is the smallest and most self-contained — a good one if it is your first PR here.

### What to do

For each handler in your slice:

[...]

Issue #39, opened 2026-07-31
Roadmap: what's open, how hard it is, and how to claim it


LeyckerS:
This is the landing page for anyone who wants to contribute. Everything open is listed below, ranked by size, with the two things that actually matter up front: **how hard it is**, and **whether you need Windows**.

MoonDownloader runs on Windows, but **almost none of the open work requires it.** Documentation, CI, tests and dependency work all run on Linux and macOS, and the test suite stubs Chrome and the network at the `moon_extract` boundary so it runs anywhere. Exactly one open issue needs a Windows machine, and it says so.

**Most of the bugs below were found by using the program**, not by reading it. Several come with measurements from real runs in the issue itself, so you can check the fix against a number instead of a feeling.

*Last checked against the repository: 29 August 2026.*

## Never contributed to open source before?

One issue is **reserved** for someone who has never had a pull request merged anywhere. It is done entirely in the browser — no install, no command line, nothing to run.

| Issue | What you change |
|:--|:--|
| #173 | Two arithmetic expressions in `web/app.js` that convert the same bytes to "GB" differently from the third one |

**#79, #80 and #93 have all been done**, by @FlaggedATX, @mazi-eth and @StefStrg, each on their first pull request and each merged without a second round. #173 is the current reserved issue and is unclaimed. If you have never had a pull request merged anywhere and #173 is taken by the time you read this, say so on this issue and I will write another — that is a standing offer, not a formality.

The work is deliberately small. The point is to get you through fork → edit → pull request once, with every step written out. If you get stuck halfway, say so on the issue — that is what it is there for.

**If you have contributed before, please leave it** and take something below.

## Free right now

Nothing here is claimed. No Windows machine needed unless the row says so.

| Issue | What it is | Size | Needs Windows |
|:--|:--|:--|:--|
| #57 | **`Engine.snapshot()` has no unit tests**, and it is the contract both front-ends read ~12 times a second. Five ready-made contract tests already exist on PR #66 and still pass on current `main` — see the 29 Aug note on the issue | ~40 lines | No |
| #178 | **A stall-killed link whose re-extraction fails is never counted** — both front-ends wait forever. The call-site analysis is in the issue; building the repro is the first half of the work | repro + a guard | No |
| #84 | The same speed is computed in MiB/s in some places and MB/s in others, all labelled "MB/s" — back in the pool 29 Aug | small, several files | No |
[...]

PR #48, opened 2026-07-31
fix: report skipped proxies and missing proxy files

### LINKED ISSUE #42: bug: --proxies fails silently when the file is missing or nothing parses
### Context

`--proxies` defaults to `proxies.txt`. When the file is absent, `ProxyPool.load()` returns early:

```python
def load(self, path: str) -> int:
    if not os.path.exists(path): return 0
```

and the caller only prints when the count is non-zero:

```python
n_proxies = _PROXY_POOL.load(proxy_path)
if n_proxies:
    print(f"[proxies] {n_proxies} loaded")
```

So a run started with `--proxies /path/that/does/not/exist` prints **nothing at all** about proxies and downloads go out directly. The user who deliberately passed a proxy file gets the one outcome they were trying to avoid, and no indication it happened.

There is a second silent path. Every malformed line is swallowed one by one:

```python
except Exception: continue
```

A proxy file in a format the parser does not recognise — anything that is not `host:port`, `host:port:user:pass`, `user:pass:host:port`, or a full `http://` / `socks` URL — loads zero proxies and still prints nothing. From the outside, a typo in the path and a file full of unparseable lines look identical to not passing the flag at all.

This matters more than a normal missing-file case, because the whole reason to pass a proxy list is to not connect directly.

### What to do

**1. Say something when the file is missing.**

If `--proxies` was given explicitly and the path does not exist, print a clear warning. If the default `proxies.txt` is simply absent, a quieter note is enough — that is the normal no-proxy case and should not look like an error.

**2. Report skipped lines.**

Have `load()` count lines it could not parse and surface that count. `12 loaded, 3 skipped` tells the user their file has a format problem; silence does not.

**3. Warn when the file exists but yields zero proxies.**

This is the case most worth catching: the file is right there, and not one line survived parsing.

**Keep the current behaviour of continuing without proxies.** This should not become a fatal error — running direct is a legitimate fallback. It just must not be silent.

### Acceptance criteria

- [ ] A non-existent explicit `--proxies` path produces a visible warning
- [ ] A file that parses to zero proxies produces a visible warning
- [ ] The skipped-line count is reported when it is greater than zero
- [ ] A missing default `proxies.txt` does not look like an error
- [ ] The run still proceeds without proxies in every case above
- [ ] A test in `tests/` covers the missing-file and zero-parsed cases

### Notes for the contributor

**No Windows needed** — `ProxyPool.load()` only touches the filesystem, and the test can build the proxy file in a `tmp_path` fixture.

Note that `moon_engine.py` carries its own copy of `ProxyPool` (see #34). Whatever you change here belongs in both until that duplication is resolved — `CONTRIBUTING.md` states this rule.

AdvaitVarhade:
### What changed
This PR fixes the silent failure modes in `--proxies` loading as described in #42. 

Specifically:
1. **Missing Files**: Added logic to emit a visible `WARNING: proxy file not found at {path}` if an explicitly passed proxy file does not exist. (If the default `proxies.txt` is missing, it still fails silently since this is the normal no-proxy state).
2. **Skipped Lines**: `ProxyPool.load()` now returns a tuple `(loaded_count, skipped_count)`. The CLI and engine both print `[proxies] {N} loaded, {S} skipped` if any lines failed to parse.
3. **Zero Proxies Loaded**: If a file exists but yields exactly 0 proxies (all lines were empty, comments, or skipped), a clear `WARNING` is printed.
4. **Shared Logic**: The exact same proxy pool loading improvements were applied to both `moon_cli.py` and `moon_engine.py` to maintain synchronization (per `CONTRIBUTING.md` rules on `ProxyPool`).
5. **Testing**: Added `tests/test_proxy_pool.py` using `pytest` and `tmp_path` to assert all warning permutations.

### Why we changed it
Previously, users could pass a misspelled path (e.g., `--proxies px.txt`) or a file with broken formatting, and the downloader would silently proceed with direct, unproxied connections. This defeated the entire security purpose of the `--proxies` flag without notifying the user. By surfacing parse counts and missing-file errors, users immediately know if their proxy setup is broken, while still preserving the fallback behavior of downloading directly.

Closes #42

### DIFF
--- a/moon_cli.py
+++ b/moon_cli.py
@@ -51,13 +51,13 @@ def _fmt_speed(mbs: float) -> str:
     return f"{mbs:.1f} MB/s" if mbs >= 1 else f"{mbs*1024:.0f} KB/s"
 
 async def run(urls: list[str], output_dir: str, n_workers: int,
-              max_dl: int, max_retries: int, proxy_path: str):
+              max_dl: int, max_retries: int, proxy_path: str, is_default_proxies: bool = False):
 
     os.makedirs(output_dir, exist_ok=True)
 
-    n_proxies = _PROXY_POOL.load(proxy_path)
-    if n_proxies:
-        print(f"[proxies] {n_proxies} loaded")
+    n_proxies, skipped = _PROXY_POOL.load(proxy_path, is_default=is_default_proxies)
+    if n_proxies or skipped:
+        print(f"[proxies] {n_proxies} loaded, {skipped} skipped")
 
     q             = asyncio.Queue()
     dl_sem        = asyncio.Semaphore(max_dl)
@@ -275,7 +275,7 @@ def main():
     ap.add_argument("--browsers", type=int, default=8,  help="Parallel extraction workers (default: 8)")
     ap.add_argument("--streams",  type=int, default=24, help="Concurrent download streams (default: 24)")
     ap.add_argument("--retries",  type=int, default=3,  help="Max retries per link (default: 3)")
-    ap.add_argument("--proxies",  default="proxies.txt", help="Proxy list file (default: proxies.txt)")
+    ap.add_argument("--proxies",  default=None, help="Proxy list file (default: proxies.txt)")
     args = ap.parse_args()
 
     if not os.path.exists(args.urls):
@@ -289,9 +289,12 @@ def main():
 
     print(f"Loaded {len(urls)} URLs from {args.urls}")
 
+    is_default_proxies = args.proxies is None
+    proxy_path = args.proxies if args.proxies is not None else "proxies.txt"
+
     try:
         asyncio.run(run(urls, args.output, args.browsers, args.streams,
-                        args.retries, args.proxies))
+                        args.retries, proxy_path, is_default_proxies))
     except KeyboardInterrupt:
         print("\nInterrupted.")
     except Exception:
--- a/moon_download.py
+++ b/moon_download.py
@@ -106,10 +106,13 @@ class ProxyPool:
         self._lock = threading.Lock()
         self._sessions: dict[str, aiohttp.ClientSession] = {}
 
-    def load(self, path: str) -> int:
+    def load(self, path: str, is_default: bool = False) -> tuple[int, int]:
         if not os.path.exists(path):
-            return 0
+            if not is_default:
+                print(f"WARNING: proxy file not found at {path}")
+            return 0, 0
         loaded = []
+        skipped = 0
         with open(path, encoding="utf-8") as f:
             for line in f:
                 line = line.strip()
@@ -132,11 +135,16 @@ class ProxyPool:
                         elif len(parts) == 2:
                             ip, port = parts
                             loaded.append({"url": f"http://{ip}:{port}", "auth": None})
+                        else:
+                            skipped += 1
                 except Exception:
                     # Skip line if proxy parsing or auth formatting fails
+                    skipped += 1
                     continue
         self.proxies = loaded
-        return len(loaded)
+        if not loaded:
+            print(f"WARNING: proxy file {path} yielded 0 proxies")
+        return len(loaded), skipped
 
     def next(self) -> dict | None:
         """Round-robin proxy selection."""
--- a/moon_engine.py
+++ b/moon_engine.py
@@ -480,7 +480,7 @@ class Engine:
             captcha_wait=eff["dn_captcha"])
 
         proxy_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "proxies.txt")
-        self._proxies = _PROXY_POOL.load(proxy_path)
+        self._proxies, skipped = _PROXY_POOL.load(proxy_path, is_default=True)
 
         n, d, r = eff["workers"], eff["dl_streams"], eff["retries"]
         self.log(f"▶  {len(urls)} links  ·  {n} extractors  ·  {d} streams  ·  {r} retries  ·  {VERSION}", "info")
@@ -489,8 +489,8 @@ class Engine:
                  f"   ·   datanodes: {applied['lanes']} pages, captcha {applied['captcha_wait']}s"
                  f"{', API key' if applied['api_key'] else ''}", "dim")
         self.log(f"   chrome: {applied['chrome']}", "dim")
-        if self._proxies:
-            self.log(f"   proxies: {self._proxies} loaded — rotating per download", "info")
+        if self._proxies or skipped:
+            self.log(f"   proxies: {self._proxies} loaded, {skipped} skipped — rotating per download", "info")
         self.log(f"   stall < {STALL_MIN_MBS} MB/s  ·  grace {STALL_GRACE_S}s  ·  max {STALL_MAX_KILL} kill", "dim")
 
         self._thread = threading.Thread(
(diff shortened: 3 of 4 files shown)


LeyckerS:
The logic is right and it covers all three cases the issue asked for — explicit path missing, file present but zero parsed, and the skipped-line count. Applying it to both `moon_cli.py` and `moon_engine.py` was correct; that two-copy rule in `CONTRIBUTING.md` catches people out.

Two things before this goes in.

**1. Trailing whitespace — 23 added lines have it.** Several are whitespace-only lines, and one is `except Exception: ` with a space after the colon. `git diff --check` flags all of them, and #26 will add `ruff` to CI, which fails on `W291`/`W293`. Worth clearing now rather than having it break a later PR.

**2. `is_default` is inferred by string comparison:**

```python
is_default = (proxy_path == "proxies.txt")
```

That treats an explicit `--proxies proxies.txt` as the default and stays silent about it, which is the one case where the user definitely asked for proxies. Cleaner is to let argparse tell you whether the flag was actually passed — set `default=None` on the argument, and resolve to `"proxies.txt"` after checking `args.proxies is None`. Then `is_default` is a fact instead of a guess.

**Heads up on a conflict, not your fault:** #41 moves `ProxyPool` out of both front-ends into a new `moon_download.py`. Whichever of the two lands second will need a rebase. Your `tests/test_proxy_pool.py` imports `from moon_cli import ProxyPool`, which becomes `from moon_download import ProxyPool` if #41 goes first. I'm intending to take yours first because it is small and self-contained, so you should be unaffected — but do not be surprised if I come back and ask.

Push the whitespace fix and the `is_default` change and I will merge.

AdvaitVarhade:
I've just pushed a commit addressing both points:
Trailing whitespace: Stripped all trailing whitespaces across the modified files. git diff --check now runs clean, so we shouldn't trip up ruff when #26 lands.
is_default logic: Updated moon_cli.py's argparse to set default=None for --proxies, and we now pass down a strict is_default_proxies = args.proxies is None boolean to determine if it was explicitly requested. The tests were also updated to match this new signature.
Everything is pushed and the tests are passing.
Please let me know if you need anything else before merging.

LeyckerS:
Both fixes check out — I verified them on `87e3d68` rather than taking it on trust:

- **Trailing whitespace**: zero added lines with trailing whitespace across all four files. `git diff --check` is clean, so `ruff` W291/W293 won't bite when #26 lands.
- **`is_default`**: the wiring is exactly right. `--proxies default=None`, `is_default_proxies = args.proxies is None`, and the engine passing `is_default=True` for its implicit `proxies.txt` — that keeps the "no proxies configured" path silent while surfacing a genuine typo'd path. `test_missing_explicit_proxy` / `test_missing_default_proxy` pin both halves of that behaviour. Good work.

Unfortunately I have to ask you for one more round, and it's my fault, not yours: **#41 landed on `main` while this was in review** and it moved `ProxyPool` out from under you.

The branch is now `mergeable_state: dirty`. On `main` today:

- `ProxyPool` lives **only** in `moon_download.py` (class at L102, `load()` at L109-138, module-level singleton `_PROXY_POOL` at L173)
- `moon_cli.py` and `moon_engine.py` no longer define it — they import `_PROXY_POOL` (`moon_cli.py` L17, `moon_engine.py` L39) and only call it (`moon_cli.py` L58, `moon_engine.py` L483)

So the rebase is mostly a simplification. What it should look like on top of `main`:

1. **`moon_download.py`** — apply the `load()` change here, once. New signature `load(self, path: str, is_default: bool = False) -> tuple[int, int]`, plus the missing-file warning, the skipped counter and the zero-parsed warning. This is the part where the diff shrinks: **the two-copy rule in `CONTRIBUTING.md` no longer applies to `ProxyPool`**, so you can drop the duplicated edit entirely. I'll update that section of `CONTRIBUTING.md` separately since it's stale now.
2. **`moon_cli.py`** — keep the argparse change (`default=None` at L278) and the `run(...)` signature/call-site change (L54-58, L294), including the `[proxies] N loaded, S skipped` print.
3. **`moon_engine.py`** — just the call site at L483: `self._proxies, skipped = _PROXY_POOL.load(proxy_path, is_default=True)`.
4. **`tests/test_proxy_pool.py`** — one import change: `from moon_download import ProxyPool` instead of `from moon_cli import ProxyPool`. The four test bodies stand as they are.
5. **`tests/test_no_chrome.py`** — the extra positional arg for the new `run()` signature stays as you have it.

Nothing conceptual to redo, and the result should be meaningfully smaller than the current diff.

One other thing worth knowing: the Lint runs on this branch are sitting at `action_required`, which is GitHub gating workflows for first-time contributors — not a failure on your side. I'm approving the run so you get real CI feedback on the rebased push, and I'll merge as soon as it's green.

Thanks for the patience on this one — the fix itself was right the first time.

LeyckerS:
One correction to my comment above, since I'd rather not leave a wrong statement standing on your PR.

I said `CONTRIBUTING.md` was stale and that I'd update it. I checked, and it isn't — #41 already updated it. It currently reads:

> **Shared logic goes in `moon_extract.py` or `moon_download.py`**, not copy-pasted between front-ends. If a change touches extraction, the Chrome lifecycle, downloading, telemetry or proxy rotation, it must land in one place and be visible from both `moon_engine.py` and `moon_cli.py`.

So the rule was always "one place", and after #41 that one place is `moon_download.py`. The document that was actually out of date was the roadmap in #39, which still described `ProxyPool` as living in two copies — that's what you'd have been reading when you scoped this work, and it's on me. I've now rewritten it.

Nothing about the rebase guidance changes: apply `load()` once in `moon_download.py`, keep the argparse and call-site changes where they are.

AdvaitVarhade:
I've just rebased bugfix/proxy-silent-fail onto the latest main following the move to moon_download.py. All 10 pytest tests pass cleanly and git diff --check is clear. Please let me know if you need anything else before merging.

LeyckerS:
Merged as `113fe80`. I checked the rebased diff rather than just the green tick, and it's right:

- `load()` applied **once**, in `moon_download.py` — signature `(path, is_default=False) -> tuple[int, int]`
- `moon_cli.py`: `--proxies` at `default=None`, `is_default_proxies = args.proxies is None`, and `run()` takes it as a defaulted parameter so nothing else calling `run()` breaks
- `moon_engine.py`: one line, `is_default=True` for the implicit `proxies.txt`
- `tests/test_proxy_pool.py` importing from `moon_download`
- Zero trailing whitespace, one commit, 62 lines against the 102 you started with

The detail I want to point out, because it's the part people usually get wrong: `if n_proxies or skipped:` rather than `if n_proxies:`. A file that parsed nothing but skipped 12 lines now prints `0 loaded, 12 skipped` instead of staying silent. That's the actual bug in #42, and you kept it correct through a rebase that touched every call site.

One thing I added myself in a follow-up commit rather than sending you round a third time: your first version had a test for the *file exists but yields zero proxies* warning, and it didn't survive the rebase. The three other paths are covered, that one wasn't. It's `test_zero_parsed_proxies` in `tests/test_proxy_pool.py` now — have a look, because it's the path the issue was really about and it's worth knowing it's pinned.

#42 closed. You're in `AUTHORS.md` for both this and #54, and this one is in the changelog under **Fixed** with the reason spelled out.

Two PRs, two merges, one clean rebase off a moving target, in a day. If you want the next thing: **#46** still has the 17 `moon_extract.py` handlers, and that's where narrowing changes what users actually see when an extraction breaks. **#55** and **#56** are also open and both came out of reviewing your work.

Issue #27, opened 2026-07-30
chore(deps): add upper bounds to requirements.txt and ship a constraints.txt


LeyckerS:
### Context

`requirements.txt` currently declares floors and no ceilings:

```
aiohttp>=3.9
playwright>=1.40
curl_cffi>=0.7        # Chrome TLS fingerprint - required for fuckingfast.co
```

Two consequences:

1. **A breaking major release installs silently.** `curl_cffi` in particular is the load-bearing dependency for `fuckingfast.co` extraction — its `impersonate` API is not stable across majors. A fresh `pip install -r requirements.txt` six months from now can pull something that does not work, and the failure surfaces as a 403 rather than as an install error.
2. **Dependabot is opening PRs against unbounded ranges.** Five have been opened so far, two of them were closed without merging — there is nothing telling the bot what the project actually supports.

There is no record of the versions the project is known to work on, so a user hitting a bug has no reference point to compare against.

### What to do

**1. Add upper bounds on the major** in `requirements.txt`:

```
aiohttp>=3.9,<4
playwright>=1.40,<2
curl_cffi>=0.7,<1     # Chrome TLS fingerprint - required for fuckingfast.co
```

Keep the existing comments — the note about `pywebview` being optional stays exactly as it is.

**2. Add `constraints.txt`** with the exact versions the project is currently known to work on. Get these from a clean install (`pip freeze`), not from memory. Include the transitive deps that matter (`certifi`, `cffi`, `greenlet`).

**3. Document it** — add the reproducible-install line to `docs/QUICKSTART.md`:

```bash
pip install -r requirements.txt -c constraints.txt
```

with one sentence explaining that `constraints.txt` pins the tested set and `requirements.txt` alone gives you the latest compatible.

### Acceptance criteria

- [ ] `pip install -r requirements.txt -c constraints.txt` completes on a clean venv
- [ ] `pip install -r requirements.txt` (no constraints) still completes
- [ ] CI green — the `no-chrome` job installs `aiohttp` and `curl_cffi` directly and must keep passing
- [ ] `docs/QUICKSTART.md` mentions the constraints file
- [ ] The version each bound is set at is justified in the PR description

### Notes for the contributor

**You do not need Windows for this.** `aiohttp` and `curl_cffi` install fine on Linux and macOS — those are the two that matter here. `playwright` is only exercised on the `datanodes.to` path.

Scope is roughly 15 lines across three files. Comment here before you start.

[...]

PR #53, opened 2026-07-31
docs: fix headless CLI example

### LINKED ISSUE #44: docs(readme): the Manual block documents a CLI invocation that does not exist
### Context

The **Manual** install block in `README.md` ends with this line:

```bash
```

That invocation does not exist. `moon_cli.py` takes no positional arguments and has no `-o` short flag. The real parser is:

```python
ap.add_argument("--urls",     required=True,  help="Text file with one URL per line")
ap.add_argument("--output",   required=True,  help="Output folder for downloaded files")
```

So the documented command fails immediately with `error: the following arguments are required: --urls, --output`, and the reader's first contact with the CLI is an argparse error.

It is also wrong in shape, not just in spelling: `--urls` takes a **path to a file containing URLs**, not URLs on the command line. Someone correcting only the flag names would still write something that does not work.

`docs/QUICKSTART.md:42` already has it right — the README is the odd one out:

```bash
python moon_cli.py --urls links.txt --output ./downloads --browsers 16 --streams 48
```

### What to do

Replace the README line with a correct minimal invocation, matching the style of the two lines above it:

```bash
python moon_cli.py --urls links.txt --output ./downloads   # headless CLI
```

Then check the rest of `README.md` for any other CLI invocation and confirm each one against the parser in `moon_cli.py`. `docs/CLI.md` (added in #37) is the reference for what is real.

### Acceptance criteria

- [ ] Every `moon_cli.py` command line in `README.md` runs without an argparse error
- [ ] No flag appears that is absent from `moon_cli.py`
- [ ] The example makes clear `--urls` is a **file**, not a URL
- [ ] No other file is touched

### Notes for the contributor

**No Windows needed.** `python moon_cli.py --help` prints the real flags on any OS — that alone is enough to verify this.

One or two lines. Comment here before you start.

tomatotomata:
Closes #44

The README manual block now uses the actual headless CLI interface:

```bash
python moon_cli.py --urls links.txt --output ./downloads
```

This matches the parser and `docs/CLI.md`: `--urls` points to a text file containing one URL per line, and `--output` names the destination directory. I checked the remaining `moon_cli.py` reference in the README and verified `python moon_cli.py --help`.

Validation: `python moon_cli.py --help`, `git diff --check`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ pip install -r requirements.txt
 playwright install chromium
 python moon_bridge.py            # GUI
 python moon_bridge.py --serve    # server only, prints the URL
-python moon_cli.py <url> ... -o <folder>   # headless CLI
+python moon_cli.py --urls links.txt --output ./downloads   # headless CLI
 ```
 
 ### Just want to look at the interface?


LeyckerS:
Checked this against the parser rather than the description, per the rule in #39 that docs claims get verified against source before merge — and it holds up:

- `moon_cli.py` L273: `--urls`, `required=True`, "Text file with one URL per line"
- `moon_cli.py` L274: `--output`, `required=True`, "Output folder for downloaded files"

So `python moon_cli.py --urls links.txt --output ./downloads` is correct, and the line it replaces (`python moon_cli.py <url> ... -o <folder>`) was wrong in two separate ways — positional URLs that the parser never accepted, and a `-o` short flag that doesn't exist. Good catch on both.

It's still marked as a draft, so I haven't merged it. **Mark it ready for review when you're happy with it and it goes in.**

One heads-up: the Lint check will sit at `action_required` because GitHub gates workflows on first-time contributor PRs. That's not a failure on your side — I approve it manually.

Also, if you want a second one after this: #27 and #26 are both unclaimed, no Windows needed for either.

LeyckerS:
Correcting myself again on this one — I told you the Lint check would sit at `action_required`. That was wrong, and in a way that's worth knowing.

`lint.yml` is filtered by `paths: ["**.py"]`, and this PR touches only `README.md`. So **no workflow runs at all** — there is nothing to approve and nothing to wait for. That's why `mergeable_state` is `clean` rather than `unstable`. Nothing is blocking you but the draft flag.

Your PR is what made me notice, so I've opened #56 to fix it: a docs-only PR currently satisfies the "CI is green" merge rule in #39 without a single check running, which for a project whose docs make checkable claims about the CLI is not good enough. The check I have in mind would extract every `python moon_cli.py ...` invocation from the Markdown and verify each flag against the parser — exactly the verification you did by hand here.

Mark it ready for review and it goes in.

LeyckerS:
@tomatotomata — this is still sitting in **draft**, which is the only thing stopping it. Everything else was settled a day ago: I verified `--urls` and `--output` against the parser, the line you replaced was wrong in two separate ways, and no workflow is blocking you.

Click **"Ready for review"** at the bottom of the PR and I merge it. If you would rather I flip it for you, say so and I will.

Worth knowing what your PR set off, because it was more than a one-line docs fix: it made me notice that documentation-only pull requests ran no checks at all, which became #56, which @Moferanoluwa fixed in #70 with a test that reads the real flag set from `moon_cli.py --help`.

And while looking into why this PR had no CI signal, I found that the new test **still passes the exact line you are fixing** — it only matches flags with two dashes, so the bogus `-o` slips through. That is now its own issue.

So the one-line fix you opened has produced a CI workflow and two issues so far. Mark it ready and let us get it in.

LeyckerS (approved):
I said I would flip this out of draft for you if you wanted, and I am doing it now rather than waiting, because the situation changed while it sat here.

Two pull requests arrived this morning for #94 — the issue about the *check* that should have caught this line — and both of them make the **byte-identical** change to `README.md` that you proposed on 31 July. They had to: you cannot tighten a test without fixing the thing it now catches, or CI goes red.

You found it first, by nine hours, and you verified it properly. Waiting for a click would have meant your fix landing as somebody else's side effect. That is not how this should work, so: merging yours.

For the record of what this one line set off — it is more than a docs fix:

- It made me notice that documentation-only pull requests ran **no checks at all**, which became #56
- @Moferanoluwa fixed that in #70 with a test that reads the real flag set from `moon_cli.py --help`
- Checking why *this* PR had no CI signal, I found that the new test still passed the exact line you were fixing, because it only matched flags with two dashes — that became #94
- Two people are now fixing #94

One line, one workflow, three issues, four contributors. Thanks for the catch — and next time, mark it ready as soon as you are happy with it.

Closes #44.

PR #54, opened 2026-07-31
refactor(cleanup): annotate exception swallows in moon_bridge.py and …

AdvaitVarhade:
This PR addresses 5 unannotated `except Exception:` handlers across `moon_bridge.py` and `moon_cli.py` as outlined in #46:
1. **`moon_bridge.py`**:
   - `native_dialog`: Narrowed `except Exception:` to `except (subprocess.TimeoutExpired, OSError):` with a comment explaining subprocess fallback behavior.
   - Tk `-topmost` setting: Added a comment explaining platform/window manager rejection.
2. **`moon_cli.py`**:
   - `ProxyPool.load`: Added comment clarifying deliberate line skipping on parsing failures.
   - `ProxyPool.close_all`: Added comment explaining best-effort cleanup on shutdown.
   - `main`: Added comment clarifying top-level CLI crash logging to `crash_log.txt`.
### Verification
- Passed all pytest tests locally
- Formatting check clean: `git diff --check` (0 issues)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,6 +36,7 @@ moon_bridge.py     loopback HTTP + token, launches Edge/Chrome --app, OS dialogs
   moon_engine.py   the engine with no GUI: start/stop/snapshot
     moon_extract.py  datanodes (real Chrome over CDP) · fuckingfast (curl_cffi)
                      BrowserGate: the launch, deferred until a datanodes link
+    moon_download.py download_file · Telemetry · ProxyPool
 
 moon_cli.py        argparse CLI, same engine, same extraction layer
 ```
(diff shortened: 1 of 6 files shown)


LeyckerS:
Reviewed — this is the right shape for #46 and I'm happy to take it as-is.

What I like about it: you picked the handlers where the *reason* for swallowing is non-obvious and documented that reason, rather than blanket-commenting all 25. The two that matter most:

- `native_dialog`: narrowing to `except (subprocess.TimeoutExpired, OSError)` is the correct set. Those are genuinely the only two ways spawning the Tk subprocess fails in a way we want to absorb — anything else (a `SyntaxError` in `_DIALOG_SRC`, say) should now surface instead of silently returning `""`. That's a real behaviour improvement, not just a comment.
- `moon_cli.py` top-level: worth stating explicitly that the handler exists to write `crash_log.txt` and re-raise the traceback, because it reads like a swallow at a glance and isn't one.

The `-topmost` and `close_all()` comments are accurate too — both are legitimately best-effort.

Also good that you targeted `moon_download.py` for the `ProxyPool` / session-cleanup handlers, since #41 consolidated those there. Saves you the conflict that #48 is currently hitting.

`mergeable: true`, so the only thing holding this is CI: the Lint run on `e899036` is at `action_required` because GitHub gates workflows on first-time contributor PRs. I'm approving the run now — assuming it comes back green, this goes in as-is with no changes requested.

If you want to keep going on #46 after this, the remaining handlers in `moon_extract.py` and `moon_engine.py` are the higher-value ones — a couple of those are hiding real extraction failures, and narrowing them properly is a more interesting piece of work than annotating. Happy to scope that out as a follow-up issue if you're interested.

LeyckerS:
Merged as `8748f17`. CI came back green on `e899036` once I approved the run, and it went in exactly as you wrote it — no changes requested, which is not the common outcome for a first PR to a repo.

You're now recorded in [AUTHORS.md](https://github.com/LeyckerS/moondownloader/blob/main/AUTHORS.md) with what you did, and the `native_dialog` narrowing is in the changelog under **Fixed** rather than as a cleanup note — it changes behaviour, and that is worth stating. A broken `_DIALOG_SRC` used to return an empty path indistinguishable from the user cancelling the dialog. It now raises.

Two things open on your side:

**#48** needs the rebase. CI passed there too, but the branch is still `dirty` against `main` — the guidance is in [my earlier comment](https://github.com/LeyckerS/moondownloader/pull/48#issuecomment-5145253319), and the short version is that `load()` applies once in `moon_download.py` now instead of twice.

**#46** if you want it: the `moon_extract.py` handlers (17 of them) are the half where narrowing actually changes what users see. Several of them currently make a broken selector look identical to a dead link. That is a more interesting piece of work than #54 was, and #55 — which I opened after your PR made me check the architecture docs — came out of the same area.

Thanks for the clean first contribution.

Issue #58, opened 2026-07-31
bug: the elapsed-time clock keeps running after a download finishes


LeyckerS:
### Context

When a run finishes, the elapsed-time clock in the GUI keeps counting up forever. It never freezes at the final duration and it never resets.

The cause is in `moon_engine.py`. `snapshot()` computes elapsed on every poll from the start timestamp:

```python
# moon_engine.py:583
el = (now - t0) if t0 else 0.0
```

`self._t0` is set in `start()` (L469) and nothing ever freezes it. When the run ends, `_on_done()` records that the run stopped but not *when* it stopped:

```python
# moon_engine.py:388-392
def _on_done(self):
    with self._lock:
        self._running = False
        self._stop_flag = False
        self._state = "done"
```

So `now - t0` keeps growing, and the GUI renders it because the stage after a run is `done`, not `idle`:

```javascript
// web/app.js:376
if (m.elapsed_s > 0 && m.stage !== "idle") parts.push(fmtClock(m.elapsed_s));
```

**The correct value already exists and is thrown away.** The run loop computes the real duration one line before calling `_on_done()`, uses it for the log line, and discards it:

```python
# moon_engine.py:383-386
el = time.monotonic()-t0; m, s = divmod(int(el), 60)
...
self.log(f"\n✓  Done in {m}m {s}s  ·  ✓ {ok}  ✗ {fail}  ⚡ {kills} kills", "ok")
self._on_done()
```

So the log tells the truth and the counter does not. This also affects the stop path, which reaches `_on_done()` the same way.

### What to do

1. Add `self._t_end = 0.0` next to `self._t0` in `__init__` (around L117).
2. Reset it in `start()` where `_t0` is set (L469), so a second run starts clean.
3. Record it in `_on_done()`, inside the existing lock.
4. In `snapshot()`, freeze the value once the run has ended — read `_t_end` inside the `with self._lock` block alongside `t0`, then compute elapsed against it when it is set.

Keep it to that. Do not change the GUI: the engine ships the number, `web/app.js` only formats it. Do not change the log line at L385.

### Acceptance criteria

- [ ] After a run completes, two snapshots taken seconds apart report the **same** `elapsed_s`
- [ ] The frozen value matches the duration in the `✓ Done in Xm Ys` log line
- [ ] Starting a second run resets the clock rather than continuing from the previous total
- [ ] Stopping a run mid-way also freezes the clock
- [ ] `elapsed_s` is still live and increasing while a run is in progress
[...]

PR #61, opened 2026-07-31
Add --version flag to moon_cli.py, record version in CLI/engine reports

### LINKED ISSUE #60: cli: no --version flag, and the report files do not record which build wrote them
### Context

There is no way to ask this project what version it is.

`moon_engine.py` defines a `VERSION` constant and writes it into the run log. `moon_cli.py` never mentions it: `python moon_cli.py --help` lists six flags and none of them is `--version`, and the CLI's own report files (`moontech_cli_*.log` / `.json`, written by `Telemetry._save_cli` in `moon_download.py:245`) do not record which build produced them.

That matters for a tool whose failure mode is external. When someone opens an issue saying extraction stopped working for a provider, the first question is which version they were on — and right now neither they nor their log files can answer it. The report file is exactly where that answer belongs.

### What to do

1. **Add `--version` to `moon_cli.py`.** Use argparse's built-in `action="version"` rather than hand-rolling a print-and-exit; it belongs with the other `add_argument` calls around L273-278. Import `VERSION` from where it already lives instead of introducing a second copy — two version constants that can disagree is a worse bug than having none.
2. **Record the version in the CLI report header.** `Telemetry._save_cli` writes a header block starting at `moon_download.py:259`. Add the version there, in the same style as the existing lines. The JSON report should carry it as a field.
3. **Do the same for the engine report** (`_save_engine`) so the GUI's reports match the CLI's.

One source of truth for the version. If that means moving the constant so both `moon_cli.py` and `moon_download.py` can import it without a circular import, say so in the PR and propose where — that decision is worth discussing before you write it.

### Acceptance criteria

- [ ] `python moon_cli.py --version` prints the version and exits 0
- [ ] It works without `--urls` and `--output`, which are otherwise required
- [ ] The version appears in the `.log` report header and as a field in the `.json` report
- [ ] Exactly one version constant exists in the codebase
- [ ] `pytest tests/ -q` still passes
- [ ] `docs/CLI.md` documents the new flag

### Notes for the contributor

**No Windows needed.** `python moon_cli.py --help` and `--version` both print and exit.

The interesting part of this issue is not the flag, it is not duplicating the constant. `argparse` makes the flag itself three lines.

Moferanoluwa:
Fixes #60 (commented there before starting, per CONTRIBUTING.md).

## What changed

1. **`moon_cli.py`**: added `--version` via `argparse`'s built-in `action="version"`, importing the existing `VERSION` constant from `moon_download.py` (no second copy). Since `action="version"` short-circuits before argparse's required-argument check, it works without `--urls`/`--output`.
2. **`moon_download.py`**: `Telemetry._save_cli` now includes `VERSION` in its log header and as a `"version"` field in its JSON report (previously had neither). `_save_engine`'s log header already had it; added the matching `"version"` field to its JSON report for parity.
3. **`docs/CLI.md`**: documented the new flag in the arguments table.

One version constant (`moon_download.py:23`), imported everywhere it's used — no circular-import issue since `moon_cli.py` already imports from `moon_download`.

## Verification

- `pytest tests/ -q` — 11 passed.
- `python moon_cli.py --version` → prints `v2.0`, exit 0, with and without other args.
- `python moon_cli.py` (no args) → still correctly errors requiring `--urls`/`--output`, exit 2 (no regression).
- Direct smoke test of `Telemetry.save()` for both `flavor="cli"` and `flavor="engine"`, confirming the generated `.log` header and `.json` field actually contain the version (not just eyeballing the diff):
  ```
  CLI json version field: v2.0
  CLI log header line: MOONTECH CLI v2.0  --  PERFORMANCE LOG
  Engine json version field: v2.0
  Engine log header line: MOONTECH v2.0  —  PERFORMANCE LOG
  ```

All acceptance criteria from the issue are met.

## AI Usage

Implemented with Claude Code (Claude Sonnet). Scope: the three files above only; no changes to extraction, download, or GUI logic.

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -21,6 +21,7 @@ are ignored.
 | `--streams` | integer | `24` | Maximum number of concurrent download streams. |
 | `--retries` | integer | `3` | Maximum extraction attempts per URL. Network retries inside one download are separate. |
 | `--proxies` | path | `proxies.txt` | Proxy-list file to load. Missing files simply result in no proxies being loaded. |
+| `--version` | flag | — | Print the version and exit. Works even without `--urls`/`--output`. |
 
 `--urls` and `--output` are required. The parser accepts integer values for
 `--browsers`, `--streams`, and `--retries`; it does not add further CLI-side range
--- a/moon_cli.py
+++ b/moon_cli.py
@@ -14,6 +14,7 @@ from urllib.parse import urlparse, unquote
 from moon_download import (
     LAUNCH_ARGS,
     Telemetry,
+    VERSION,
     _PROXY_POOL,
     _close_sess,
     _sanitize_filename,
@@ -276,6 +277,7 @@ def main():
     ap.add_argument("--streams",  type=int, default=24, help="Concurrent download streams (default: 24)")
     ap.add_argument("--retries",  type=int, default=3,  help="Max retries per link (default: 3)")
     ap.add_argument("--proxies",  default=None, help="Proxy list file (default: proxies.txt)")
+    ap.add_argument("--version",  action="version", version=VERSION)
     args = ap.parse_args()
 
     if not os.path.exists(args.urls):
--- a/moon_download.py
+++ b/moon_download.py
@@ -267,7 +267,7 @@ class Telemetry:
             buf.write(" ".join(str(p) for p in parts) + "\n")
 
         W("=" * 72)
-        W("MOONTECH CLI  --  PERFORMANCE LOG")
+        W(f"MOONTECH CLI {VERSION}  --  PERFORMANCE LOG")
         W("=" * 72)
         W(f"Duration : {int(el//60)}m {int(el%60)}s")
         if ok_r:
@@ -291,6 +291,7 @@ class Telemetry:
         )
         with open(jp, "w", encoding="utf-8") as f:
             json.dump({
+                "version": VERSION,
                 "duration_s": round(el, 2),
                 "ok": len(ok_r),
                 "fail": len(recs) - len(ok_r),
@@ -388,6 +389,7 @@ class Telemetry:
         with open(jp, "w", encoding="utf-8") as f:
             json.dump({
                 "session": {
+                    "version": VERSION,
                     "start": datetime.datetime.now().isoformat(),
                     "duration_s": round(el, 2),
                     "config": self.cfg,


LeyckerS:
Merged as `b0e881e`. Welcome to the project.

I checked every claim in your description against the source rather than taking the diff at face value, and all of it holds:

- **One `VERSION` constant.** `moon_download.py:23` is the only definition in the codebase. `moon_engine.py` imports it (L37), `moon_bridge.py` re-imports it from there, and now `moon_cli.py` does too. That was the part of #60 most likely to go wrong and you got it right.
- **`_save_engine`'s log header already carried it** — `moon_download.py:319`, `MOONTECH {VERSION}`. Your description said so, and it was true. Adding the matching JSON field for parity was the right call rather than leaving the two report flavours asymmetric.
- **`action="version"` short-circuits the required-argument check.** Correct, and it is why this works without `--urls`/`--output` — which is exactly what the acceptance criteria asked for.
- Zero trailing whitespace, three files, six added lines, nothing outside the stated scope.

Two things I want to single out, because they are the difference between a PR I merge in five minutes and one that costs an afternoon.

**You smoke-tested the actual output.** Printing the generated `.log` header and the `.json` field, and pasting them into the description, means I did not have to guess whether the f-string landed in the right place. Most contributors assert the diff looks right; you showed the artifact.

**You disclosed the AI usage and scoped it.** That is the correct way to do it, and it does not count against the PR here. What matters is whether the claims are verifiable and the scope held — both did.

One cosmetic thing I decided **not** to send you back for: `version=VERSION` prints `v2.0` rather than `moon_cli.py v2.0`, where argparse convention would be `version=f"%(prog)s {VERSION}"`. The issue asked for the version and exit 0, you delivered that, and I am not spending your time on a nitpick that was not in the acceptance criteria. Mentioning it only so you know it was considered rather than missed.

You are now in [AUTHORS.md](https://github.com/LeyckerS/moondownloader/blob/main/AUTHORS.md) with what you did, and this is in the changelog under **Changed**.

If you want another: your own PR surfaced a stale row in `docs/CLI.md` — the `--proxies` documentation still describes the pre-#48 behaviour. That is now #62, and it is yours if you want it since you already have the file open. **#58** is also open and is the better one if you would rather touch Python: a real bug where the elapsed clock never stops after a run, with a three-line fix.

Issue #56, opened 2026-07-31
ci: documentation-only pull requests run no checks at all


LeyckerS:
### Context

`.github/workflows/lint.yml` is filtered by path:

```yaml
on:
  push:
    branches: [main]
    paths:
      - "**.py"
      - ".github/workflows/lint.yml"
  pull_request:
    paths:
      - "**.py"
```

So a pull request that touches only `.md` files triggers **no workflow at all**. Not a skipped job, not a neutral check — nothing runs.

That matters more here than in most projects, because documentation is the largest category of contribution to this repository and the failure mode is specific: **the docs make checkable claims about the CLI.** #44 existed because `README.md` documented `python moon_cli.py <url> ... -o <folder>`, an invocation with a positional argument the parser never accepted and a `-o` flag that does not exist. That line sat in the README through every green build, because no build ever looked at it.

#39 tells contributors that "CI is green" is a merge requirement. For a docs-only PR that requirement is currently vacuous, and the only thing standing between a wrong command and the README is me reading the diff carefully.

### What to do

Add a job that verifies the CLI invocations in the Markdown against the actual parser.

A reasonable shape, but propose your own if you see a better one:

1. New workflow (or a new job in `lint.yml`) triggered on `paths: ["**.md"]` as well as `**.py`.
2. Extract every `python moon_cli.py ...` command from the tracked `.md` files.
3. For each one, check that every `--flag` it uses exists in `moon_cli.py`'s parser. Reading the flags out of `moon_cli.py --help` is the reliable way — do not hardcode a list, or the check goes stale exactly like the docs did.
4. Fail with a message that names the file, the line and the offending flag.

The check must not need a browser, a display or Playwright. `moon_cli.py --help` prints and exits, so this stays a fast pure-Python job.

**Deliberately out of scope:** validating prose, spelling, or link checking. This is about commands that do not work, nothing more.

### Acceptance criteria

- [ ] A documentation-only pull request produces at least one running check
- [ ] The job fails if a `.md` file uses a `moon_cli.py` flag absent from the parser
- [ ] The job passes on `main` as it stands today
- [ ] The flag list is derived from `moon_cli.py`, not hardcoded in the workflow
- [ ] Adding a real new flag to `moon_cli.py` does not require editing the workflow
- [ ] The job needs no browser and no Playwright install
- [ ] `no-chrome` and `syntax-check` still pass unchanged

### Notes for the contributor

[...]

PR #63, opened 2026-07-31
docs(cli): fix stale --proxies row and document the actual warning behavior

### LINKED ISSUE #62: docs(cli): the --proxies row still documents the pre-#48 silent behaviour
### Context

#48 changed how `--proxies` behaves. `docs/CLI.md` still documents the old behaviour, and #61 happened to put the stale line right next to a diff, which is how I noticed.

**`docs/CLI.md:23` currently says:**

> \| `--proxies` \| path \| `proxies.txt` \| Proxy-list file to load. Missing files simply result in no proxies being loaded. \|

Two things in that row are now wrong:

**1. "Missing files simply result in no proxies being loaded" is no longer true.** That sentence describes exactly the bug #42 reported and #48 fixed. Today:

- an **explicitly passed** path that does not exist prints `WARNING: proxy file not found at {path}`
- a file that exists but parses to nothing prints `WARNING: proxy file {path} yielded 0 proxies`
- unparseable lines are counted and reported as `[proxies] N loaded, S skipped`
- only the **implicit** `proxies.txt` stays silent when absent, because that is the normal no-proxy state

The distinction between those last two cases is the entire point of the fix, and the docs currently describe the opposite behaviour.

**2. The default column says `proxies.txt`, but the parser's default is now `None`** (`moon_cli.py:278`). The effective fallback is still `proxies.txt` (`moon_cli.py:293`), so `proxies.txt` is the right thing to show a user — but if the table is meant to mirror the parser, the difference is worth a word, because it is what makes the silent-versus-warning behaviour possible.

There is a third row at `docs/CLI.md:113` in the GUI-comparison table — check whether it needs anything, but it reads fine to me as-is.

### What to do

Rewrite the `--proxies` row at L23 so it describes the current behaviour. Keep it to the width and tone of the neighbouring rows: this is a reference table, not a changelog. If the warning behaviour needs more than one sentence, the table row can stay short and the detail can go in the prose below it — your judgement.

Then check the rest of `docs/CLI.md` for any other claim that #48, #54 or #61 made stale. Verify each one against `moon_cli.py` and `moon_download.py` before you write it down; do not trust this issue as a complete list.

### Acceptance criteria

- [ ] No sentence in `docs/CLI.md` states that a missing proxy file is silent
- [ ] The three warning cases and the silent default case are distinguishable from the documentation
- [ ] Every default shown in the table matches the parser, or the difference is explained
- [ ] Only `docs/CLI.md` is modified

### Notes for the contributor

**No Windows needed and nothing to install.** Everything you need is `moon_cli.py` L276-297 and `ProxyPool.load()` in `moon_download.py` L109-147. `python moon_cli.py --help` prints the real flags.

The skill this issue rewards is the one that matters most in a docs PR here: reading the code to find out what actually happens, instead of writing what the flag sounds like it should do. That is how #44 was fixed and it is why that PR was merged without a second round.

Moferanoluwa:
Fixes #62 (commented there before starting, per CONTRIBUTING.md).

## What changed

Only `docs/CLI.md`. Rewrote the `--proxies` row (was describing pre-#48 behavior) and added a short prose block distinguishing the four actual cases, verified directly against `ProxyPool.load()` (`moon_download.py:109-147`) and its call site in `run()` (`moon_cli.py:59-61`) rather than taking the issue's summary at face value:

- an explicitly passed path that doesn't exist → `WARNING: proxy file not found at {path}`
- a file that exists but yields no usable proxies → `WARNING: proxy file {path} yielded 0 proxies` (this one fires regardless of default vs. explicit)
- the implicit default (`proxies.txt`) missing → silent, the normal no-proxy state
- whenever anything loads or gets skipped → `[proxies] {n} loaded, {s} skipped`

Also noted that the parser's actual default for `--proxies` is `None` (`moon_cli.py:278`), with `proxies.txt` applied as a fallback afterward (`moon_cli.py:294-295`) — that's what makes the silent-vs-warning distinction possible in the first place, so the table's shown default isn't literally what the parser has.

## Checked, not just the one row

- The GUI-comparison table row at L113 (`--proxies` / — / "The CLI reads a proxy-list file; it is not a GUI panel setting.") — still accurate, makes no claim about warning behavior. Left as-is.
- What #48 and #54 actually touched: #48 changed `moon_cli.py`, `moon_download.py`, and `moon_engine.py` (the proxy-warning behavior itself); #54 was comment-only exception annotations, no behavior change. Nothing else in `docs/CLI.md` was affected by either.

## AI Usage

Implemented with Claude Code (Claude Sonnet). Scope: `docs/CLI.md` only, as the issue's acceptance criteria required.

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -20,12 +20,22 @@ are ignored.
 | `--browsers` | integer | `8` | Number of parallel extraction workers. Despite the name, it does not start that many browsers: datanodes uses one shared Chrome instance. |
 | `--streams` | integer | `24` | Maximum number of concurrent download streams. |
 | `--retries` | integer | `3` | Maximum extraction attempts per URL. Network retries inside one download are separate. |
-| `--proxies` | path | `proxies.txt` | Proxy-list file to load. Missing files simply result in no proxies being loaded. |
+| `--proxies` | path | `proxies.txt` | Proxy-list file to load. A missing file, or one that yields no usable proxies, prints a warning — except the implicit default (`proxies.txt` when `--proxies` is omitted), whose absence stays silent as the normal no-proxy state. |
 | `--version` | flag | — | Print the version and exit. Works even without `--urls`/`--output`. |
 
 `--urls` and `--output` are required. The parser accepts integer values for
 `--browsers`, `--streams`, and `--retries`; it does not add further CLI-side range
 validation.
+
+The parser's actual default for `--proxies` is unset (`None`); `proxies.txt` is the
+effective fallback path applied afterward. That distinction is what lets the CLI tell
+"you didn't ask for proxies" (silent) apart from "you asked for a proxy file and it's
+missing or empty":
+
+- an explicitly passed path that doesn't exist: `WARNING: proxy file not found at {path}`
+- a file that exists but yields no usable proxies: `WARNING: proxy file {path} yielded 0 proxies`
+- the implicit default path missing: no warning
+- whenever any proxies load or lines get skipped: `[proxies] {n} loaded, {s} skipped`
 
 ## Examples
 


LeyckerS:
Merged as `b2a7c58`. Second one today, and this is the better of the two.

The part worth calling out is a detail **my own issue got wrong**. #62 listed four cases and implied the implicit default stays quiet in general. You went and read `ProxyPool.load()` instead of trusting my summary, and caught that the zero-yield warning is not gated:

```python
# moon_download.py:145-146
if not loaded:
    print(f"WARNING: proxy file {path} yielded 0 proxies")
```

No `is_default` check. So an implicit `proxies.txt` that exists but parses to nothing **does** warn — only the missing-file branch is silenced. Your prose says exactly that, and it is now the only place in the repo where that distinction is written down correctly. I verified it against the source before merging, as I did with your #61.

The rest holds up too:

- `[proxies] {n} loaded, {s} skipped` fires on `if n_proxies or skipped` (`moon_cli.py:60`) — your wording matches the condition, including the skipped-only case
- The `None` default with `proxies.txt` as the fallback (`moon_cli.py:278`, L293-295) is stated without making the table lie to a reader who just wants the effective value
- You checked the L113 GUI-comparison row and left it alone because it makes no claim about warnings. Correct call, and saying so in the PR meant I did not have to re-check it myself
- You verified what #48 and #54 actually touched rather than assuming my issue listed everything

One thing you could not have known, and it is the reason this merged with no checks at all: `lint.yml` filters on `paths: ["**.py"]`, so a documentation-only PR triggers **no workflow**. Nothing was skipped or pending — nothing ran. That is #56, still open, and your two PRs are now the second and third piece of evidence for why it matters: the docs here make checkable claims about the CLI, and today the only thing checking them is me reading the diff.

You are the first contributor to come back for a second PR the same day since the project opened up. `AUTHORS.md` and the changelog are updated.

If you want a third, **#58** is the one I would pick: a real bug where the elapsed clock never stops after a run finishes, three lines of state, and the correct value already exists in the code and is thrown away. **#56** is also yours if you want to close the loop on the CI gap you have now hit twice.

Issue #45, opened 2026-07-31
cli: --browsers does not control browsers — rename to --extractors with a deprecated alias


LeyckerS:
### Context

The CLI flag is called `--browsers`. It does not control browsers.

```python
ap.add_argument("--browsers", type=int, default=8, help="Parallel extraction workers (default: 8)")
...
asyncio.run(run(urls, args.output, args.browsers, args.streams, args.retries, args.proxies))

async def run(urls: list[str], output_dir: str, n_workers: int, ...):
```

It is the extraction worker count. The number of browsers is decided entirely by `BrowserGate` in `moon_extract.py`: a `fuckingfast.co` batch opens **zero** regardless of this value, and a batch containing a `datanodes.to` link opens **exactly one shared instance**, also regardless of this value.

So `--browsers 16` does not open sixteen browsers. It never opens more than one.

The rest of the project already uses the right word. The parameter is `n_workers`, the CLI's own startup line prints `{n_workers} extractors`, the engine logs `extractors`, and the GUI slider is labelled **Extractors**. `--browsers` is the only place that says browsers, and it is the surface users actually type.

### This is not hypothetical

A contributor writing the CLI documentation in #40 read the flag name and wrote:

> `--browsers` controls how many headless browser instances run in parallel

Both halves wrong — there is one browser, and it is not headless, because Turnstile issues no token to a headless build. That PR was closed, but the mistake came straight from the name. `CONTRIBUTING.md` already says names should describe intent; this one describes something that does not happen.

### What to do

Add `--extractors` as the real flag and keep `--browsers` working as a deprecated alias.

**Do not just rename it.** People have this in scripts and scheduled jobs, and a hard rename breaks them silently at the worst moment.

```python
ap.add_argument("--extractors", type=int, default=8,
                help="Parallel extraction workers (default: 8)")
ap.add_argument("--browsers", type=int, default=None,
                help=argparse.SUPPRESS)   # deprecated alias for --extractors
```

Then, after parsing: if `--browsers` was given, use it, and print a one-line deprecation notice naming the replacement. If both were given, prefer `--extractors` and say so.

Hiding the old flag from `--help` with `argparse.SUPPRESS` keeps it working without advertising it to new users.

**Also update the docs** — `docs/CLI.md`, `docs/QUICKSTART.md:42` and any README example — to use `--extractors`, with one line noting `--browsers` is the deprecated alias.

### Acceptance criteria

- [ ] `--extractors N` works and is what `--help` shows
[...]

PR #70, opened 2026-08-01
ci: verify Markdown-documented moon_cli.py flags actually exist

### LINKED ISSUE #56: ci: documentation-only pull requests run no checks at all

Moferanoluwa:
Fixes #56 (commented there before starting, per CONTRIBUTING.md).

## The gap, and why it's not hypothetical

`lint.yml` is filtered on `paths: ["**.py"]`, so a documentation-only PR triggers no workflow at all — not a skipped job, nothing runs. #44 existed because README.md documented a `moon_cli.py` invocation the parser never accepted, and it survived every green build because no build ever looked at it. My own #63 and #69 are now further evidence: `LeyckerS` had to catch the stale content by reading each diff personally, since nothing else was checking.

## What this adds

- **`tests/test_docs_cli_flags.py`**: extracts every `python moon_cli.py ...` invocation from tracked `.md` files (joining `\`-continued multi-line commands into one logical command first), and checks every `--flag` each one uses against the real flag set read from `python moon_cli.py --help` at test time. Not hardcoded — adding a real flag to `moon_cli.py` never requires touching this file.
- **New workflow `.github/workflows/docs-cli-check.yml`**, triggered on `**.md` as well as `**.py`. This is a separate file from `lint.yml` rather than a new job inside it, because GitHub Actions path filters apply to a workflow's trigger as a whole, not per-job — folding this into `lint.yml` would have made `syntax-check`/`ruff`/`no-chrome` also re-run on every doc-only change, which isn't what those jobs are for.

## Verified for real, twice over

Same lesson as #69: this PR's own diff wouldn't self-test against `pull_request`'s path filters in the way that matters, so I pushed to my fork's `main` first and read real Actions runs rather than trusting local simulation:

- **Docs CLI Check**: https://github.com/Moferanoluwa/moondownloader/actions/runs/30678079789 — `docs-cli-flags` green.
- **Lint** (triggered too, since the new test file is itself `.py`): https://github.com/Moferanoluwa/moondownloader/actions/runs/30678079768 — `no-chrome` and all three `syntax-check` matrix legs (3.10/3.11/3.12, including the `ruff check .` step from #69) green.

Also validated the failure path locally before any of that, per the issue's own suggestion: appended a deliberately nonexistent `--totally-fake-flag` to `docs/CLI.md`, ran the test, got

```
docs/CLI.md:132: '--totally-fake-flag' is not a real moon_cli.py flag
```

then reverted it. Full local `pytest tests/ -q` is 12/12 (up from 11 — this file is auto-discovered, so it also runs for free inside `no-chrome` on any future `.py` change, on top of the dedicated workflow).

## AI Usage

Implemented with Claude Code (Claude Sonnet). Scope: the two new files listed above only.

### DIFF
--- a/.github/workflows/docs-cli-check.yml
+++ b/.github/workflows/docs-cli-check.yml
@@ -1,0 +1,31 @@
+name: Docs CLI Check
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "**.md"
+      - "**.py"
+      - ".github/workflows/docs-cli-check.yml"
+  pull_request:
+    paths:
+      - "**.md"
+      - "**.py"
+
+jobs:
+  docs-cli-flags:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.10"
+
+      - name: Install engine dependencies
+        run: pip install "aiohttp>=3.9" "curl_cffi>=0.7" "pytest==8.*"
+
+      - name: Verify documented moon_cli.py flags exist
+        run: pytest tests/test_docs_cli_flags.py -q
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,6 +13,10 @@ on:
 jobs:
   syntax-check:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.10", "3.11", "3.12"]
     steps:
       - name: Checkout
         uses: actions/checkout@v7
@@ -20,7 +24,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          python-version: "3.10"
+          python-version: ${{ matrix.python-version }}
 
       - name: Byte-compile every module
         run: |
@@ -29,6 +33,12 @@ jobs:
             moon_extract.py moon_bridge.py render_gui.py \
             integration_http.py integration_web.py \
             tests/conftest.py tests/test_no_chrome.py
+
+      - name: Install ruff
+        run: pip install "ruff==0.16.1"
+
+      - name: Lint with ruff
+        run: ruff check .
 
   no-chrome:
     runs-on: ubuntu-latest
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -23,7 +23,8 @@ of truth for the modern engine. `moon_engine.py` is that engine, standing on its
 
 The async engine itself was **not** rewritten. `_run`, `_browser_worker`, `_do_dl`,
 `download_file`, `Telemetry` and `ProxyPool` are the 14.8 code, plus the lazy browser
-launch described below.
+launch described below. `download_file`, `Telemetry` and `ProxyPool` now live in
+`moon_download.py`, shared by `moon_engine.py` and `moon_cli.py`.
 
 ## Startup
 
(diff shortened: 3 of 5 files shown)


LeyckerS (approved):
This is the best-designed of the four. Reading the real flag set from `moon_cli.py --help` at test time instead of hardcoding it means adding a flag never requires touching this file to keep it honest — that is the difference between a check that survives and one that rots into a liar within two releases. The backslash-continuation handling matters too: the README block in #44 was exactly that shape.

It also closes the loop the other PRs kept demonstrating. #68 and #69 both ran **zero** checks because `lint.yml` only triggers on `**.py`; this workflow triggers on `**.md`, so a docs-only PR is finally verified against the parser. Closes #56.

One thing to know before #45 lands, since it will bite there and not here: #45 plans to rename `--browsers` to `--extractors` **keeping a deprecated alias**. If that alias is hidden with `help=argparse.SUPPRESS`, it will not appear in `--help`, so `_real_flags()` will not contain it — and any documentation that still mentions `--browsers` will fail this test even though the flag genuinely works. Not a change to make now; just worth knowing the answer in advance (either don't suppress it, or let the test read `parser._actions` rather than `--help` text).

Separately, `lint.yml`'s `compileall` list still stops at `tests/test_no_chrome.py`, so neither `tests/test_proxy_pool.py` nor your new `tests/test_docs_cli_flags.py` is byte-compiled. Not yours to fix here — I am opening it as its own issue.

PR #71, opened 2026-08-01
cleanup: annotate the 3 except Exception: handlers in moon_engine.py

Moferanoluwa:
Slice D of #46 (commented there before starting, per CONTRIBUTING.md) — the `moon_engine.py` portion; `moon_extract.py`'s slices A/B/C are @AdvaitVarhade's per the latest count in the issue.

## The three handlers, per the #54 standard

- **L265** (per-URL extraction/scheduling inside the worker loop) — stays broad, with a comment. Any failure processing one URL (network, parsing, Chrome/CDP) must not kill the worker loop, which is still serving the rest of the queue.
- **L371** (writing the telemetry report at run end) — narrowed to `(OSError, TypeError)`. `OSError` covers file open/write failures (permissions, disk full, bad path); `TypeError` covers `json.dump` choking on a non-serializable field. The run's downloads have already completed by this point, so logging a warning instead of propagating is correct — just not for literally any exception.
- **L505** (`_guarded_run`'s top-level try around `asyncio.run` in a thread) — stays broad, with a comment. The function's own docstring already explains why ("an escaping exception would vanish silently"); added the same reasoning at the `except` clause itself to match how #54 annotated its broad catches inline, not just at the function level.

## Verification

- `python -m py_compile moon_engine.py` — clean.
- `ruff check moon_engine.py` — clean, no new findings from the `#69` baseline.
- `pytest tests/ -q` — 11/11, no regression. Checked first that nothing in `tests/` exercises `telem.save`'s exception path, so narrowing that one carried no test-breakage risk.

## AI Usage

Implemented with Claude Code (Claude Sonnet). Scope: `moon_engine.py` only, the three handlers listed in the issue's slice D.

### DIFF
--- a/moon_engine.py
+++ b/moon_engine.py
@@ -263,6 +263,8 @@ class Engine:
                             async with tasks_lock: all_tasks.append(t)
                             success = True
                 except Exception as e:
+                    # One URL's extraction/scheduling failed for any reason (network,
+                    # parsing, Chrome/CDP); the worker loop must keep serving the queue.
                     rec.notes.append(f"exception: {e}")
                     self.log(f"    ✗  {e}", "fail")
 
@@ -368,7 +370,10 @@ class Engine:
             lp, jp = telem.save(base)
             self.log(f"📊  {os.path.basename(lp)}", "info")
             self.log(f"📊  {os.path.basename(jp)}", "info")
-        except Exception as e:
+        except (OSError, TypeError) as e:
+            # OSError: can't open/write the report files (permissions, disk full, bad
+            # path). TypeError: json.dump choking on a non-serializable field. The run's
+            # downloads already completed; a report-write failure shouldn't crash finalize.
             self.log(f"⚠  Log save error: {e}", "warn")
 
         if output_links and mode == "links":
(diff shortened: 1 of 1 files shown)


Moferanoluwa:
Force-pushed a correction: the branch was accidentally cut from a local main that included content from my still-open #68 and #69 (an artifact of testing those on my own fork's main to get real Actions runs, per the approach in their PR descriptions). Diff is now scoped to just moon_engine.py, cherry-picked cleanly onto actual upstream main. Sorry for the noise.

LeyckerS (approved):
Slice D, and done the way the bar asked.

I checked the narrowing rather than taking it on trust, because `except (OSError, TypeError)` at the end of a successful run is the one change here that could bite:

- `Telemetry._save_cli` / `_save_engine` open both files with an explicit `encoding="utf-8"`, so `UnicodeEncodeError` — the obvious escape on a Windows box handling arbitrary host-supplied filenames — cannot occur.
- The division sites (`tb/el/1e6`) can't raise `ZeroDivisionError`: `t0` and `t_end` both come from `time.monotonic()`, so `el` is never exactly zero.
- `os.makedirs` and both `open()` calls give `OSError`; `json.dump` on a non-serializable field gives `TypeError`. Both covered.

So the set is right, and your comment names the actual reason for each half rather than "just in case".

The other two correctly stay broad. The `_guarded_run` one in particular has to — it is the last net under a background thread, and your comment says exactly that.

PR #73, opened 2026-08-01
fix: freeze elapsed-time clock when a run finishes

### LINKED ISSUE #58: bug: the elapsed-time clock keeps running after a download finishes

RubenSanosh:
Closes #58

   Adds `_t_end` alongside `_t0`: recorded in `_on_done()`, reset in `start()`,
   and used in `snapshot()` to freeze `elapsed_s` once a run has ended instead
   of letting it keep counting from `now - t0`.

   Added tests/test_elapsed_clock.py covering:
   - elapsed_s stays frozen after a run finishes
   - elapsed_s resets when a new run starts
   - elapsed_s still increases while a run is in progress

   Verified with `python -m pytest tests/ -q` — 14 passed.

### DIFF
--- a/moon_engine.py
+++ b/moon_engine.py
@@ -115,6 +115,7 @@ class Engine:
         self._dls        = 0
         self._bytes_acc  : collections.deque = collections.deque(maxlen=200000)
         self._t0         = 0.0
+        self._t_end      = 0.0
         self._proxies    = 0
 
         # Live FileRecord registry: the GUI reads these objects every snapshot,
@@ -390,6 +391,7 @@ class Engine:
             self._running = False
             self._stop_flag = False
             self._state = "done"
+            self._t_end = time.monotonic()
 
     def scan_tmp(self) -> int:
         folder = self._cfg["out_folder"]
@@ -467,6 +469,7 @@ class Engine:
             self._ok = 0; self._fail = 0; self._kills = 0
             self._browsers = 0; self._dls = 0
             self._bytes_acc.clear(); self._t0 = time.monotonic()
+            self._t_end = 0.0
             self._tracked.clear()
         with self._log_lock:
             self._log_ring.clear(); self._log_total = 0
@@ -558,6 +561,7 @@ class Engine:
             state    = self._state
             running  = self._running
             t0       = self._t0
+            t_end    = self._t_end
             url_done = self._url_done; url_tot = self._url_total
             dl_done  = self._dl_done;  dl_tot  = self._dl_total
             ok       = self._ok; fail = self._fail
@@ -580,7 +584,7 @@ class Engine:
         else:
             eta = 0.0
 
-        el = (now - t0) if t0 else 0.0
+        el = (t_end - t0) if t_end else ((now - t0) if t0 else 0.0)
         # No phase sentence here on purpose: the GUI owns wording and language,
         # so the engine ships numbers and a stage name instead of prose.
         if not running and state == "idle":
--- a/tests/test_elapsed_clock.py
+++ b/tests/test_elapsed_clock.py
@@ -1,0 +1,42 @@
+import time
+from moon_engine import Engine
+
+
+def test_elapsed_freezes_after_done():
+    engine = Engine()
+    engine._t0 = time.monotonic()
+    engine._on_done()
+
+    first = engine.snapshot()["metrics"]["elapsed_s"]
+    time.sleep(0.2)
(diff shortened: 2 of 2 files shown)


LeyckerS (approved):
Correct, and correct for the right reason. Merging.

The part I checked rather than assumed is whether `_on_done()` is really the only way a run ends — a second exit path would leave the clock running in exactly the case nobody tests. It is: `_running = False` and `_state = "done"` appear only inside `_on_done()` itself (the other occurrences are `__init__`), and `_on_done()` has exactly two callers — the tail of `_run` and the `except` in `_guarded_run`. So both the normal finish and the crash path freeze the clock. Nothing is left uncovered.

The expression handles all four states cleanly:

| State | `t_end` | `t0` | Result |
|:--|:--|:--|:--|
| never run | 0 | 0 | `0.0` |
| running | 0 | set | `now - t0`, live |
| finished | set | set | `t_end - t0`, frozen |
| restarted | reset to 0 | reset | live again |

Resetting `_t_end` in `start()` is the half that is easy to forget, and you did it in the same locked block as the other counters, so a snapshot can never observe a new `_t0` with a stale `_t_end`.

Three tests covering frozen / reset / still-live is the right set — the third is the one that stops a future "fix" from freezing the clock permanently.

One note for next time, not a change to this PR: `test_elapsed_resets_on_new_run` calls the real `start()`, which runs `_moon_extract.configure()` and loads `proxies.txt` from disk before your `_guarded_run` stub is reached. It passes and it is not wrong, but those two are real side effects in a unit test and worth stubbing if you touch this file again.

Nice first contribution here — thanks for taking it and for saying up front it might take you a little longer.

PR #77, opened 2026-08-01
ci: derive compileall file list from git ls-files

### LINKED ISSUE #74: ci: the compileall list is hand-maintained and has drifted — moon_download.py is not byte-compiled
`.github/workflows/lint.yml` byte-compiles an explicit, hand-written list of files:

```yaml
python -m compileall -q \
  moon_engine.py moon_cli.py \
  moon_extract.py moon_bridge.py render_gui.py \
  integration_http.py integration_web.py \
  tests/conftest.py tests/test_no_chrome.py
```

Every file added since that list was written is missing from it. Right now that is:

| Missing | Why it matters |
|:--|:--|
| `moon_download.py` | **The shared download engine.** `download_file`, `Telemetry` and `ProxyPool` all live here since #41, and both front-ends import from it. It is the single most-depended-on module in the repo and nothing compiles it |
| `tests/test_proxy_pool.py` | added later, never compiled |
| `tests/test_docs_cli_flags.py` | added in #70, never compiled |

`moon_download.py` is the one that matters. A syntax error in it would still be caught by the `no-chrome` job at import time, so this is not a hole you can drive a broken build through — but the job named "Byte-compile every module" does not compile every module, and it will silently fall further behind with each new file.

### The fix

Don't repair the list — derive it, so it cannot drift again:

```yaml
- name: Byte-compile every module
  run: git ls-files '*.py' | xargs python -m compileall -q
```

`git ls-files` only returns tracked files, so nothing in `__pycache__`, a virtualenv or an untracked scratch file gets pulled in.

### Acceptance criteria

- `lint.yml` no longer contains a hand-written list of `.py` files
- The job passes on all three matrix versions
- Deliberately introduce a syntax error in `moon_download.py` locally, confirm the new command fails, then revert it — say in the PR that you did

Small, no Windows needed, and it touches one file.

darlenepolek:
Closes #74

## What
`lint.yml` byte-compiled an explicit hand-maintained list of `.py` files, which drifted — `moon_download.py` (the most-depended-on module: `download_file`, `Telemetry`, `ProxyPool`), `tests/test_proxy_pool.py` and `tests/test_docs_cli_flags.py` were never compiled.

Replace the list with a derived one so it can never drift again:

```yaml
run: git ls-files '*.py' | xargs python -m compileall -q
```

Only tracked files are included — no `__pycache__`, venv or scratch files.

## Verification (acceptance criteria)
- `lint.yml` no longer contains a hand-written list of `.py` files ✅
- New command compiles all 13 tracked `.py` files, exit 0 ✅
- Per the issue: I deliberately introduced a syntax error into `moon_download.py` locally — the new command **fails** (exit 123), while the old hand-maintained list **passed** (exit 0), confirming the hole. Reverted before committing.
- Job runs on the existing 3-version matrix (3.10/3.11/3.12) unchanged.

One file touched: `.github/workflows/lint.yml` (+1 −6).
- Bonus: added \`.github/workflows/lint.yml\` to the \`pull_request\` path filter — otherwise this PR (workflow-only change) would never trigger CI, and the "job passes on all three matrix versions" criterion could not be verified.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -9,6 +9,7 @@ on:
   pull_request:
     paths:
       - "**.py"
+      - ".github/workflows/lint.yml"
 
 jobs:
   syntax-check:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -27,12 +27,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Byte-compile every module
-        run: |
-          python -m compileall -q \
-            moon_engine.py moon_cli.py \
-            moon_extract.py moon_bridge.py render_gui.py \
-            integration_http.py integration_web.py \
-            tests/conftest.py tests/test_no_chrome.py
+        run: git ls-files '*.py' | xargs python -m compileall -q
 
       - name: Install ruff
         run: pip install "ruff==0.16.1"


LeyckerS (approved):
Taking this one. It arrived first — 03:21 against 05:37 for #87, which proposes the same `git ls-files` change — and it is also the better patch, for a reason worth spelling out.

You added `.github/workflows/lint.yml` to the `pull_request` paths trigger. #87 did not, and the consequence is visible in the two pull requests themselves: **yours ran the full matrix and came back green, theirs ran no checks at all.** A workflow that does not watch itself cannot be verified by the change that edits it. That was a second, unstated bug in the same file, and you fixed it in the same breath.

The core change is right too. `git ls-files '*.py'` only returns tracked files, so nothing from `__pycache__`, a virtualenv or an untracked scratch file gets pulled in, and the list can never drift again. `moon_download.py` — the shared download engine that nothing was compiling — is now covered, along with `tests/test_proxy_pool.py` and `tests/test_docs_cli_flags.py`.

Closes #74. Thanks, and welcome.

PR #82, opened 2026-08-01
refactor(extract): narrow and document exception handlers in moon_extract.py (Slice A)

AdvaitVarhade:
Resolves part of #46 (Slice A: `moon_extract.py` lines 71–556).

### Changes Made
Audited and refactored the 8 exception handlers in `moon_extract.py` (lines 71–556):

1. **Line 71 (`curl_cffi` optional import):** Narrowed `except Exception:` to `except ImportError:`.
2. **Line 101 (`close_ff_session`):** Added `# best-effort cleanup during shutdown` comment for deliberate cleanup swallow.
3. **Line 188 (`ff_extract` retry loop):** Added `# network or parsing error triggers a retry` comment.
4. **Line 378 (`prepare_datanodes_context`):** Locally imported `PlaywrightError` and narrowed to `except PlaywrightError:`.
5. **Line 391 (`_dn_eval`):** Locally imported `PlaywrightError` and narrowed to `except PlaywrightError as e:`.
6. **Line 471 (`dn_api_extract`):** Added `# swallow network/JSON errors and fall back to scraping` comment.
7. **Line 537 (`_dn_click_turnstile` auto-click):** Locally imported `PlaywrightError` and `PlaywrightTimeoutError` and narrowed to `except (PlaywrightError, PlaywrightTimeoutError):`.
8. **Line 556 (`_dn_click_turnstile` manual click fallback):** Narrowed to `except (PlaywrightError, PlaywrightTimeoutError) as e:`.

*Note:* Playwright exception types were imported locally inside the respective functions to preserve the module's deferred lazy-loading of Playwright.

### Verification
- `pytest` passed (15/15 tests passing, including `test_no_chrome.py`).
- `git diff --check` passed cleanly.

### DIFF
--- a/moon_extract.py
+++ b/moon_extract.py
@@ -68,7 +68,7 @@ FALLBACK_UA    = ("Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 
 try:
     from curl_cffi import AsyncSession as _CurlSession
     HAVE_CURL_CFFI = True
-except Exception:                                    # pragma: no cover
+except ImportError:                                  # pragma: no cover
     _CurlSession = None
     HAVE_CURL_CFFI = False
 
@@ -99,7 +99,7 @@ async def close_ff_session():
         try:
             await _FF_SESSION.close()
         except Exception:
-            pass
+            pass  # best-effort cleanup during shutdown
         _FF_SESSION = None
 
 
@@ -186,6 +186,7 @@ async def extract_fuckingfast(url: str) -> str | None:
                 return m.group()
             _d(f"ff {file_id}: status={status} no hx-redirect")
         except Exception as e:
+            # network or parsing error triggers a retry
             _d(f"ff {file_id}: {type(e).__name__} {e}")
 
         if attempt + 1 < FF_RETRIES:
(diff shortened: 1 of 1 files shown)


LeyckerS (approved):
Slice A, and the hard half of #46. Merging.

The thing I checked hardest was the function-level `from playwright.async_api import Error`, because importing a heavy optional dependency inside a function usually means someone did not know where imports go. Here it is the opposite: `moon_extract.py` deliberately has **no** module-level playwright import, and the comment at line 1107 says why — *"its own function so a batch that never touches datanodes never imports"* it. Hoisting those three imports to the top would have quietly destroyed the property that `tests/test_no_chrome.py` exists to protect. You kept it, and the `no-chrome` job is green, which is the proof.

On the narrowings:

- `except ImportError` around the `curl_cffi` import is right. CPython raises `ImportError` for a failed extension-module load too, so a missing native runtime still lands in the same branch and `HAVE_CURL_CFFI` still ends up `False` instead of taking the whole module down.
- `_dn_eval` narrowing to `PlaywrightError` matches its own docstring — "Execution context was destroyed" is a Playwright error and nothing else was ever going to reach that string check.
- The three that stay broad each say what they are now instead of nothing.

Two small things, neither blocking:

**1. The tuple in `_dn_click_turnstile` is redundant.** `playwright.async_api.TimeoutError` subclasses `Error`, so `except (PlaywrightError, PlaywrightTimeoutError)` catches exactly what `except PlaywrightError` alone would. Harmless, and arguably clearer as documentation of intent — but worth knowing it is not doing extra work.

**2. `close_ff_session`'s comment is the weakest of the six.** "best-effort cleanup during shutdown" describes the situation rather than stating a fact about what can go wrong. The bar #54 set is a specific reason — something closer to *"the session may already be closed or its transport gone; nothing downstream depends on this succeeding."*

Neither is worth another round trip. Slices B and C are still open if you want them.

PR #87, opened 2026-08-01
fix(ci): derive compileall file list instead of hardcoding it

XEDAB:
Fix the #74 issue: change the old compile instruction in `lint.yml` that misses files to one that iterates over all `.py` files. I broke `moon_download` and confirmed it does error (and I've restored `moon_download` to its original state).

Compile error:
```
*** Error compiling 'moon_download.py'...
  File "moon_download.py", line 24
    def 关注塔菲喵()：
               ^
SyntaxError: invalid character '：' (U+FF1A)
```
Exit code: 123

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -27,12 +27,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Byte-compile every module
-        run: |
-          python -m compileall -q \
-            moon_engine.py moon_cli.py \
-            moon_extract.py moon_bridge.py render_gui.py \
-            integration_http.py integration_web.py \
-            tests/conftest.py tests/test_no_chrome.py
+        run: git ls-files '*.py' | xargs python -m compileall -q
 
       - name: Install ruff
         run: pip install "ruff==0.16.1"


LeyckerS:
@XEDAB — closing this one, and I am sorry, because this is the second time in a day you have been beaten to a claim.

#77 landed the same `git ls-files` change two hours earlier, and it also added `lint.yml` to the workflow's own `paths` trigger. That difference is visible right here: their pull request ran the full matrix and came back green, **this one ran no checks at all**, because the workflow does not currently watch itself. Same idea, one fix further.

Nothing wrong with your patch. You were second, and my rule is first PR wins — otherwise I am guessing, and guessing is worse.

So: **#90 is yours if you want it, and I will hold it for you.** Nobody else has claimed it and I am not giving it to anyone else for now.

It is a real bug found by running the program. Interrupting and resuming works on one provider and silently restarts from zero on the other, because when the server answers `200` instead of `206` the code discards the partial file without a word. The issue has the exact lines. The minimum fix is one log line so the user knows why they lost 4 GB; the better fix is establishing whether it can be avoided at all.

It is a more interesting problem than either of the two you missed, and it is not going anywhere. Say the word on #90 and it is yours.

XEDAB:
@LeyckerS Well... looks like I'm late again, sorry qwq
I'll take #90.

PR #98, opened 2026-08-01
fix: log when a server ignores the resume request

### LINKED ISSUE #90: bug: when the server ignores a Range request the partial file is silently discarded
Reported from real runs: interrupting and restarting **works on fuckingfast** and **does not work on datanodes** — the transfer starts again from zero, with nothing said about why.

## Cause

`moon_download.py:466-476`:

```python
if r.status not in (200, 206):
    return False, f"HTTP {r.status}", resume
if r.status == 200 and resume > 0:
    resume = 0
...
f = open(tmp, "ab" if resume > 0 else "wb")
```

The resume logic itself is correct: the size of the `.tmp` becomes `Range: bytes=N-`, and a `206` appends. But when the server answers **`200`** — "here is the whole file, I ignored your Range" — the code zeroes `resume`, which then makes `open(..., "wb")` **truncate the partial file**, and moves on. There is no log line, no warning, no telemetry note.

From the user's side a 4 GB partial file vanishes and a 4 GB download starts again, and the program looks like resume is simply broken.

Why the two providers differ is the part worth establishing: a datanodes direct link is freshly extracted on every run and may be a new signed URL that will not honour a Range against a previous session's bytes, whereas a fuckingfast link resolves to a stable CDN object that returns `206`. Whoever takes this should confirm that against a real capture rather than assume it.

## What is needed

At minimum the discard must stop being silent:

```
⚠  <name>: server ignored the resume request (HTTP 200) — restarting from zero
```

Then decide whether it can be avoided at all:

- If the fresh link would honour a Range and the old bytes are still valid, retry the request against the newly extracted URL before giving up on the partial
- If the host genuinely never supports it, say so once per provider instead of per file, and consider not writing a `.tmp` that can never be resumed

## Acceptance criteria

- Interrupting and restarting a datanodes download logs, in plain words, why it started over
- The line appears in the GUI log and in `moontech_*.log`
- fuckingfast resume keeps working unchanged — it does today, and there is no reason for it to regress
- Say in the pull request what the server actually returned in your test

Found by running the program.

XEDAB:
Fix #90: log when the server ignores a resume request.

When a server answers 200 instead of 206 during a resumable download, moon_download.py:468 resets resume to 0, which makes the .tmp file get truncated - and nothing said why. This adds a rec.notes entry so the restart is visible in the GUI log and moontech_*.log.

Verification: new regression test tests/test_resume_200.py fakes a 200 response and asserts the note is recorded. pytest tests/ -q -> 16 passed.
图片

### DIFF
--- a/moon_download.py
+++ b/moon_download.py
@@ -467,6 +467,7 @@ async def download_file(
                     return False, f"HTTP {r.status}", resume
                 if r.status == 200 and resume > 0:
                     resume = 0
+                    rec.notes.append(f"{rec.filename}: server ignored the resume request (HTTP 200) — restarting from zero")
 
                 file_size = int(r.headers.get("Content-Length", 0)) + resume
                 if file_size > 0:
--- a/tests/test_resume_200.py
+++ b/tests/test_resume_200.py
@@ -1,0 +1,91 @@
+"""Test: when a server ignores a resume request (HTTP 200), a note is logged.
+
+Regression test for #90. The resume branch at moon_download.py:468 resets
+`resume` to 0 when the server answers 200, which truncates the partial file.
+This test fakes a 200 response and asserts the note is recorded.
+"""
+from __future__ import annotations
+
+import asyncio
+import collections
+import os
+
+import moon_download
+from moon_download import FileRecord, download_file
+
+
+
+# download_file does `async with dl_session.get(url) as r:` then reads
+# `r.status`, `r.headers` and `r.content.iter_chunked(...)`. These three
+# fakes implement exactly that surface, no network involved.
+
+
+class FakeChunkIter:
+    """Empty async iterator for r.content.iter_chunked()."""
+
+    def __aiter__(self):
+        return self
+
+    async def __anext__(self):
+        raise StopAsyncIteration
+
+
+class FakeContent:
+    def iter_chunked(self, size):
+        return FakeChunkIter()
+
+
+class FakeResponse:
+    def __init__(self, status: int, headers: dict):
+        self.status = status
+        self.headers = headers
+        self.content = FakeContent()
+
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self, *exc):
+        return False
+
+
+class FakeSession:
+    def __init__(self, status: int, headers: dict):
+        self._status = status
+        self._headers = headers
+
+    def get(self, *args, **kwargs):
+        return FakeResponse(self._status, self._headers)
+
+
+def _make_tmp(tmp_path, size: int) -> str:
+    """Create a partial file so resume > 0, return the dest path."""
+    dest = str(tmp_path / "file.bin")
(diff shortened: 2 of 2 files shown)


LeyckerS (approved):
Merging. Third time was the charm — thanks for coming back after two collisions.

**The test is the best part of this PR.** `FakeSession` / `FakeResponse` / `FakeContent` implement exactly the three things `download_file` touches — `r.status`, `r.headers`, `r.content.iter_chunked()` — and nothing else. No network, no sleep, no mocking library, and it will not break the next time an unrelated part of aiohttp changes. Writing a 100-byte `.tmp` to force `resume > 0` is precisely the right way to reach that branch. This is the pattern I would point other people at.

**An acceptance criterion I got wrong, and it is my mistake not yours.** #90 says the message must appear "in the GUI log and in `moontech_*.log`". You did the second. The first is not reachable from where you were working: `rec.notes` surfaces only at `moon_download.py:362`, inside report writing, and `download_file` lives in the shared module with no access to `Engine.log()` — the engine logs live, `download_file` records to the record. I wrote a criterion that assumed a channel the architecture does not have.

Your fix is the right one for where the code is today, and it turns a silent 4 GB loss into something the user can actually find. I have opened a separate issue for the live-log channel, because that is a design change and not something this PR should have carried.

One cosmetic thing, not worth a round trip: `f"{rec.filename}: server ignored..."` repeats the filename, and `rec.notes` is already per-record — the report prints it in a row that names the file. Harmless duplication.

Closes #90.

XEDAB:
> Merging. Third time was the charm — thanks for coming back after two collisions.
> 
> **The test is the best part of this PR.** `FakeSession` / `FakeResponse` / `FakeContent` implement exactly the three things `download_file` touches — `r.status`, `r.headers`, `r.content.iter_chunked()` — and nothing else. No network, no sleep, no mocking library, and it will not break the next time an unrelated part of aiohttp changes. Writing a 100-byte `.tmp` to force `resume > 0` is precisely the right way to reach that branch. This is the pattern I would point other people at.
> 
> **An acceptance criterion I got wrong, and it is my mistake not yours.** #90 says the message must appear "in the GUI log and in `moontech_*.log`". You did the second. The first is not reachable from where you were working: `rec.notes` surfaces only at `moon_download.py:362`, inside report writing, and `download_file` lives in the shared module with no access to `Engine.log()` — the engine logs live, `download_file` records to the record. I wrote a criterion that assumed a channel the architecture does not have.
> 
> Your fix is the right one for where the code is today, and it turns a silent 4 GB loss into something the user can actually find. I have opened a separate issue for the live-log channel, because that is a design change and not something this PR should have carried.
> 
> One cosmetic thing, not worth a round trip: `f"{rec.filename}: server ignored..."` repeats the filename, and `rec.notes` is already per-record — the report prints it in a row that names the file. Harmless duplication.
> 
> Closes #90.

thank for your review

130658046_p0_master1200

Issue #96, opened 2026-08-01
ci: the project promises Python 3.10+ but tests only 3.10-3.12 — including the version the maintainer actually runs


LeyckerS:
The README badge, the requirements section and `docs/QUICKSTART.md` all promise **Python 3.10+**. That is open-ended: it covers 3.13 and 3.14, both released.

`.github/workflows/lint.yml` tests:

```yaml
python-version: ["3.10", "3.11", "3.12"]
```

So two of the versions the project claims to support are never exercised. The lower bound is verified and the upper end is not verified at all.

**The maintainer of this repository develops on Python 3.14.6.** That is the practical problem, and it cuts both ways:

- A bug found while using the program may be specific to 3.14 and unreproducible for a contributor on 3.10, who then cannot fix what they cannot see.
- Green CI says nothing about the environment the program is actually being run in day to day.

## What is needed

Add `"3.13"` and `"3.14"` to the matrix.

The interesting part is what happens next. `fail-fast: false` is already set, so a failure on a new version reports without hiding the others. If something does break, **that is the finding**, and it should become its own issue rather than being papered over — say in the pull request what you saw on each version.

If everything passes, the change is three characters and the promise in the README is finally backed by something.

While you are there, `ruff.toml` pins `target-version = "py310"`. That is correct and should stay: it means "the syntax must remain valid on the oldest supported version", which is not the same question as "does it run on the newest". Do not change it to match the matrix.

## Acceptance criteria

- The matrix covers every version the documentation promises
- All five report a result — no version silently skipped
- If a version fails, the pull request says exactly how, and a separate issue is opened for the failure
- `ruff.toml`'s `target-version` is untouched

One file, no Windows machine needed.

Vam-si-krish:
I’m preparing a focused PR for #96 adding Python 3.13 and 3.14 to .github/workflows/lint.yml. I’ll leave ruff.toml unchanged and validate the workflow plus the local lint/test suite.

Issue #95, opened 2026-08-01
an unsupported host is retried like a failure and reported as one, with no message saying it is unsupported


LeyckerS:
Paste a link from any host the program does not handle — a YouTube URL, anything — and it is reported as `failed`, indistinguishable from a dead datanodes link. Found by running the program.

## Cause

`_browser_worker` in `moon_engine.py` dispatches on the host:

```python
if "fuckingfast.co" in parsed.netloc:
    ...
elif "datanodes.to" in parsed.netloc:
    ...
```

**There is no `else`.** An unsupported host matches neither branch, so nothing runs, nothing is logged, and `success` is still `False` when control reaches the retry block:

```python
if not success and not is_re and attempt < max_retries and not self._get("_stop_flag"):
    backoff = min(2**(attempt-1), 6)
    self.log(f"    ↻  retry in {backoff}s", "warn")
    await asyncio.sleep(backoff)
    await q.put((url, attempt+1, rec))
```

So the URL is **requeued and retried**. With the default `retries = 3` that is two extra attempts and roughly three seconds of sleep, each holding a worker, for a host that can never succeed no matter how many times it is tried. Retrying is the right response to a network blip; it is the wrong response to "I do not support this website".

The user-facing half is worse than the wasted work: the only lines that reach the log are `↻ retry in 1s` and then a failure. Nothing says why. Paste a mixed list and the ten failures give no way to tell which links are dead and which are sites the program was never going to handle.

## What is needed

An explicit `else` branch that ends the attempt immediately:

- Do not requeue — an unsupported host is terminal on the first look
- Log something unambiguous, e.g. `✗ unsupported host: youtube.com — supported: datanodes.to, fuckingfast.co`
- Put a distinguishable note on the record (`rec.notes`) so `moontech_*.log` separates "unsupported" from "extraction failed"
- Name the supported hosts from one place rather than hardcoding the strings a third time — they are already spelled out twice in this function and again in `classify()` in `web/app.js`

Worth considering, but say which you chose: the GUI's link editor already colours unsupported links differently via `classify()`, so this could also be caught before a run starts rather than during it. Catching it at paste time is friendlier; catching it in the engine also covers the CLI. Doing it in the engine is the one that must exist.

## Acceptance criteria

- An unsupported URL is marked failed on the **first** attempt, with no retry and no backoff sleep
- The log says the host is unsupported and names what is supported
- `moontech_*.log` distinguishes it from a genuine extraction failure
[...]

PR #104, opened 2026-08-01
test: catch short flags in CLI docs

### LINKED ISSUE #94: test_docs_cli_flags misses short flags — it passes the exact line that motivated it
`tests/test_docs_cli_flags.py` (added in #70, closing #56) verifies that every `moon_cli.py` invocation in the documentation uses flags the parser accepts. It was built because of #44, where `README.md` documented a command the parser never accepted.

**It does not catch the line from #44.**

`README.md:147` today:

```
```

That line is wrong in two ways: `moon_cli.py` takes no positional URLs, and there is no `-o` short flag — the real ones are `--urls` and `--output`, both `required=True` (`moon_cli.py:273-274`).

The test passes it anyway, because of this:

```python
FLAG_RE = re.compile(r"--[a-zA-Z][a-zA-Z-]*")
```

Two dashes are required. `-o` has one, so `FLAG_RE.findall(...)` returns an empty list for that line, the loop over `used_flags - real_flags` has nothing to iterate, and the check reports success.

So the guard that exists to prevent this class of mistake currently certifies the original instance of it as correct. #53 (@tomatotomata) fixes the line itself; this issue is about the check.

## What is needed

Extend the check to short flags, and validate them against the parser's real short options rather than a hardcoded list — same principle that makes the long-flag half self-maintaining.

Two things to be careful about, because a naive `-[a-zA-Z]` pattern will produce false positives:

- `--flag` also matches a single-dash pattern unless you anchor it. Match short flags only when not preceded by another dash.
- Bare `-` and things like `-1` appear in ordinary prose and URLs. Restrict to tokens inside a fenced block on a line that already contains both `python` and `moon_cli.py`, which the test already does.

Also worth considering while you are in there: the test only flags **unknown** flags. It cannot tell that `moon_cli.py <url>` passes a positional argument the parser rejects. Catching that is a bigger change — say in the pull request whether you attempted it or deliberately left it, either answer is fine.

## Acceptance criteria

- With `README.md:147` in its current state, `pytest tests/test_docs_cli_flags.py` **fails** and names the line
- After #53 merges (or with the line corrected locally), it passes
- No false positives across the tracked Markdown files as they stand — run it and say so

Found while checking why #53 had no CI signal.

kocaemre:
## Description

Closes #94.

Extends the Markdown CLI docs guard so it checks documented short flags against the real `moon_cli.py --help` output too. The current README example used `-o`, which argparse does not accept; I updated that example to the long flags so the stricter guard is green on `main` even before #53 lands.

I did not add positional-argument validation in this PR; this stays scoped to the short-flag gap described in #94.

## Type of change

- [x] Documentation update
- [ ] Refactor / code cleanup
- [ ] Other:

## Checklist

      applied the equivalent change to `moon_cli.py` (not applicable; docs/test only)
      description

## Test evidence

RED:

- `python -m pytest tests/test_docs_cli_flags.py -q` failed as expected after broadening the flag regex: `README.md:147: '-o' is not a real moon_cli.py flag`

GREEN:

- `python -m pytest tests/test_docs_cli_flags.py -q` → 1 passed
- `python -m pytest tests/ -q` → 16 passed
- `git diff --check` → passed

Not applicable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ pip install -r requirements.txt
 playwright install chromium
 python moon_bridge.py            # GUI
 python moon_bridge.py --serve    # server only, prints the URL
-python moon_cli.py <url> ... -o <folder>   # headless CLI
+python moon_cli.py --urls links.txt --output ./downloads   # headless CLI
 ```
 
 ### Just want to look at the interface?
--- a/tests/test_docs_cli_flags.py
+++ b/tests/test_docs_cli_flags.py
@@ -16,7 +16,7 @@ import sys
 ROOT = pathlib.Path(__file__).resolve().parents[1]
 
 FENCE_RE = re.compile(r"```(?:bash|sh|text)?\n(.*?)```", re.DOTALL)
-FLAG_RE = re.compile(r"--[a-zA-Z][a-zA-Z-]*")
+FLAG_RE = re.compile(r"(?<!-)--[a-zA-Z][a-zA-Z-]*|(?<!-)-[a-zA-Z](?![a-zA-Z-])")
 
 
 def _real_flags() -> set[str]:


LeyckerS (approved):
Taking this one. It arrived 31 minutes before #105, which proposes the same change, and my rule is first pull request wins — I applied it against someone else earlier today and I am not going to bend it now.

The regex is correct. I ran both proposals over every tracked Markdown file in the repo and they produce **identical results on every line that exists today** — so this is not a case of merging the weaker option.

Your `README.md` change is byte-identical to #53, which @tomatotomata opened on 31 July and which I merged just before this. That is not a criticism: tightening the check without fixing the line it now catches would have turned CI red, so you had no choice. It just means that half of your diff has already landed and git resolved it silently, as it does when both sides make the same change.

Closes #94. Thanks — and #96 and #95 are both open and unclaimed if you want another.

PR #105, opened 2026-08-01
test: validate short CLI flags in Markdown docs (closes #94)

### LINKED ISSUE #94: test_docs_cli_flags misses short flags — it passes the exact line that motivated it

HeaTTap:
## Summary

Extends `tests/test_docs_cli_flags.py` to validate short CLI options (single-dash flags like `-o`) against the parser's actual flag set output from `moon_cli.py --help`.

- Updated `FLAG_RE` in `tests/test_docs_cli_flags.py` using negative lookbehind `r"(?<![\w-])--?[a-zA-Z][a-zA-Z-]*"` to match both short and long flags without false positives.
- Corrected line 147 of `README.md` which contained invalid CLI invocation syntax (`-o`).

Closes #94.

## Verification
- Verified `pytest tests/test_docs_cli_flags.py` fails on `README.md:147` before correction with `AssertionError: README.md:147: '-o' is not a real moon_cli.py flag`.
- Verified all 16 unit tests pass after updating `README.md:147`.
- Verified `ruff check` passes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ pip install -r requirements.txt
 playwright install chromium
 python moon_bridge.py            # GUI
 python moon_bridge.py --serve    # server only, prints the URL
-python moon_cli.py <url> ... -o <folder>   # headless CLI
+python moon_cli.py --urls links.txt --output ./downloads   # headless CLI
 ```
 
 ### Just want to look at the interface?
--- a/tests/test_docs_cli_flags.py
+++ b/tests/test_docs_cli_flags.py
@@ -16,7 +16,7 @@ import sys
 ROOT = pathlib.Path(__file__).resolve().parents[1]
 
 FENCE_RE = re.compile(r"```(?:bash|sh|text)?\n(.*?)```", re.DOTALL)
-FLAG_RE = re.compile(r"--[a-zA-Z][a-zA-Z-]*")
+FLAG_RE = re.compile(r"(?<![\w-])--?[a-zA-Z][a-zA-Z-]*")
 
 
 def _real_flags() -> set[str]:


LeyckerS:
@HeaTTap — closing this, and I want to be precise about why, because **your regex is the better one** and I am not merging it.

#104 arrived 31 minutes earlier with the same fix, and my rule here is first pull request wins. I applied that rule against another contributor earlier today; bending it now because I prefer a pattern would make it worthless.

But your version is genuinely more robust, and I checked rather than assumed. Run over every tracked Markdown file as they stand today, the two are **identical** — zero differing lines. The difference only shows on input that does not exist yet:

```
python moon_cli.py --output ./my-x
   #104 -> ['--output', '-x']      <- false positive
   yours -> ['--output']
```

`(?<!-)` only refuses a preceding dash. Yours, `(?<![\w-])`, refuses a preceding word character too, so a hyphen inside a path or a filename cannot be mistaken for a short flag. The first time somebody documents a path like that, the merged version fails the build and blames a flag that was never there.

**So: would you open that as its own pull request?** Change `FLAG_RE` to your pattern, and add a test case with a path like `./my-x` so the reason is recorded and nobody reverts it later. It is a small diff and a real hardening, and it is yours — I am not giving it to anyone else.

You lost this one on the clock, not on the merits. Say the word and I will open the issue for it, or just send the PR.

PR #108, opened 2026-08-01
fix: fail unsupported hosts without retrying

### LINKED ISSUE #95: an unsupported host is retried like a failure and reported as one, with no message saying it is unsupported

Vam-si-krish:
## Description

Fail unsupported hosts on their first dispatch instead of retrying them. Supported host names now live in the shared extraction module, both engine and CLI dispatchers use them, and engine telemetry records a distinct `unsupported host` note.

The guard deliberately remains at dispatch time rather than duplicating it in the GUI paste handler. The GUI already colors unsupported links, while engine and CLI guards provide the required terminal behavior for every run path.

Closes #95

## Type of change

- [ ] Documentation update
- [ ] Refactor / code cleanup
- [ ] Other:

## Checklist

N/A — no UI change. Validation: `python3 -m pytest` (18 passed) and Ruff on all touched Python files.

AI-assisted contribution; I reviewed the queue, retry, logging, telemetry, and no-browser paths.

### DIFF
--- a/moon_cli.py
+++ b/moon_cli.py
@@ -32,6 +32,9 @@ from moon_extract import (                       # noqa: E402
     HAVE_CURL_CFFI,
     DN_API_KEY,
     DN_LANES,
+    DATANODES_HOST,
+    FUCKINGFAST_HOST,
+    SUPPORTED_HOSTS,
 )
 
 if DN_API_KEY:
@@ -172,9 +175,10 @@ async def run(urls: list[str], output_dir: str, n_workers: int,
             print(f"  -> {filename[:60]}{suffix}")
 
             success = False
+            unsupported = False
             try:
                 parsed = urlparse(url)
-                if "fuckingfast.co" in parsed.netloc:
+                if FUCKINGFAST_HOST in parsed.netloc:
                     link = await extract_fuckingfast(url)
                     rec.extract_s = time.monotonic() - t_start
                     if not link:
@@ -186,7 +190,7 @@ async def run(urls: list[str], output_dir: str, n_workers: int,
                         t = asyncio.create_task(_task())
                         async with tasks_lock: all_tasks.append(t)
                         success = True
-                elif "datanodes.to" in parsed.netloc:
+                elif DATANODES_HOST in parsed.netloc:
                     # API key set -> single JSON GET, no browser, no captcha.
                     # get_browser() is where Chrome is actually launched, so a
                     # batch with no datanodes link never opens one. After that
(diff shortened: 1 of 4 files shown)


LeyckerS (approved):
Merging. This is the most complete pull request the repository has had.

**The test is what makes it.** Running with `retries=3` and then asserting `not any("retry in" in message ...)` proves the actual fix rather than the symptom — a test that only checked the failure count would still pass if the URL were retried three times and failed at the end, which is exactly the behaviour being removed. You parameterised `run_engine`/`run_cli` to make that assertion possible instead of hardcoding around it, and you wrote it for **both** front-ends. That is the standard here and very few people hit it first time.

**The deduplication went further than the issue asked.** I flagged the two dispatchers; you also routed `FF_HOST` and `DN_ALWAYS_ALLOW` through the same constants, so `"fuckingfast.co"` and `"datanodes.to"` now appear once each in the Python tree instead of four times. That is the half that stops this drifting again.

`parsed.hostname or parsed.netloc or "(missing host)"` is the right chain — a malformed URL still produces a readable message instead of an empty one.

**On leaving the GUI paste-time check alone:** correct call, and thank you for stating it rather than leaving it ambiguous. `classify()` in `web/app.js` cannot import Python constants, so a fourth copy there is unavoidable for now; putting the terminal behaviour in the dispatcher means every path gets it, including the CLI, which has no paste step at all.

Noted on the AI-assisted disclosure. It makes no difference to the review — the code is judged on what it does — but saying so unprompted is the right instinct and I would rather people did.

Closes #95.

One thing worth knowing, unrelated to this PR: datanodes extraction is currently broken upstream (#109 — their step 2 now needs two clicks and our trigger handler latches after the first). So if you test with a real datanodes link it will hang. fuckingfast is unaffected, and the no-Chrome suite covers both paths without touching the network.

Issue #99, opened 2026-08-01
download_file has no way to reach the live log — events only surface after the run


LeyckerS:
Split out of #90, where I wrote an acceptance criterion the architecture cannot currently satisfy.

`download_file` lives in `moon_download.py`, shared by the GUI engine and the CLI. When something noteworthy happens mid-transfer it can only call `rec.notes.append(...)`, and `rec.notes` is read in exactly one place — `moon_download.py:362`, inside report writing. So the user finds out **after the run**, from a file they may never open.

Three events already take this route and none of them reach the live log:

| Where | Note |
|:--|:--|
| `moon_download.py:553` | `connection dropped att N` |
| `moon_download.py:559` | `timeout att N` |
| `moon_download.py:566` | `error att N: ...` |
| `moon_download.py:469` | `server ignored the resume request (HTTP 200)` — added in #98 |

That last one is the clearest case for why it matters: a partial file is discarded and a multi-gigabyte transfer starts over, and nothing on screen says so while it happens.

Meanwhile `Engine.log()` exists and feeds the GUI's live log through the ring buffer and `snapshot()`, and the CLI prints its own lines — but neither is reachable from the shared module, and hardcoding either would break the separation that #41 established.

## What is needed

A way for `download_file` to emit an event that both front-ends can surface live. The obvious shape is an optional callback passed in alongside `telem`, defaulting to `None` so nothing breaks:

```python
async def download_file(..., telem: Telemetry | None = None, on_event=None):
    ...
    if on_event: on_event(f"⚠  {rec.filename}: server ignored the resume request — restarting", "warn")
```

`moon_engine.py` passes `self.log`; `moon_cli.py` passes something that prints. `telem.stall()` is the existing precedent for an event channel out of this module and is worth reading first — the difference is that stall events are also logged separately by the engine, which is exactly the duplication this would remove.

Whoever takes this should say in the pull request whether they also routed the three older notes through it, or deliberately limited the change to the resume one.

## Acceptance criteria

- A resume-ignored event appears in the GUI log **while the run is happening**
- The same event appears in the CLI output
- It still appears in `moontech_*.log` — the report must not lose anything
- `download_file` keeps working when no callback is passed; `tests/test_resume_200.py` passes unchanged
- No front-end-specific import lands in `moon_download.py`

Related: #97, which is the same separation problem seen from the CLI's side.

LeyckerS:
**A concrete example of the cost, from a real run today.**

[...]

PR #111, opened 2026-08-01
fix(datanodes): follow the step-2 trigger chain instead of clicking once

### LINKED ISSUE #109: datanodes extraction hangs: step 2 now needs two clicks and the trigger is a one-shot latch
**Datanodes extraction currently fails for every free account.** Every link hangs at step 2 for the full `DN_STEP2_TIMEOUT` (420 s), returns nothing, gets retried, and hangs again. No error is shown — the run simply sits at `0/N done` forever.

This is not a regression from any merged pull request. It is datanodes changing their step-2 flow, which is the failure mode `moon_extract.py`'s own header comments warn about.

## Evidence

`MOON_DEBUG=1`, ten datanodes links:

```
[extract] step 1 submitted (single POST, method_free present)
[extract] turnstile click attempt 1: no target
[extract] turnstile solved after 3.6s (1 auto-click attempt(s))
[extract] clicked step-2 trigger: free download
standard speed
```

…and then nothing for the rest of the run. Every counter stays at zero.

Note the label that was clicked: `"free download\nstandard speed"`. The button visible on the page **after** that click is a different one — `"Start Download / Your file is ready"`, under a `STEP 2 OF 2` heading.

## Cause

The trigger click is latched:

```python
if st["trigger"] and not clicked:
    await _dn_eval(page, DN_OVERLAY_JS)
    if await _dn_eval(page, DN_CLICK_JS, st["trigger"], default=False):
        _d("clicked step-2 trigger:", st["trigger"])
        clicked = True
```

`clicked` is set once and never cleared, so the loop clicks **exactly one** element per attempt no matter what the page does afterwards.

Datanodes now serves two buttons in sequence: `Free Download / Standard Speed` reveals `Start Download / Your file is ready`, and only the second one starts the transfer. The code clicks the first, latches, and spends the next 420 seconds polling a page whose new button it can see and will not press.

The detection is not the problem. `DN_STEP2_JS` already matches the new button — `start download` is the first alternative in its regex:

```js
/start download|download now|get link|proceed to download|download file|free download/
```

Every poll after the first click returns `trigger: "start download\nyour file is ready"`. The latch is the only reason it is ignored.

## The fix

Track **which label was clicked**, not merely *that* something was clicked:

```python
clicked_labels: set[str] = set()
...
if st["trigger"] and st["trigger"] not in clicked_labels:
    if await _dn_eval(page, DN_CLICK_JS, st["trigger"], default=False):
        _d("clicked step-2 trigger:", st["trigger"])
        clicked_labels.add(st["trigger"])
```

That clicks each distinct trigger once and follows a multi-step flow without looping on the same element. Points to settle in the pull request:

- **Cap it.** A page that cycles through labels would otherwise be clicked forever inside the 420 s budget. A small limit — three or four distinct triggers — with a debug line when it is hit.
- **Ordering.** `DN_STEP2_JS` breaks on the *first* DOM match, so if both buttons are ever present at once it may pick the wrong one first. Preferring `start download` over `free download` when several match would be more robust than relying on document order.
- **Do not just add a blind second click.** The number of steps is datanodes' to change again; the fix has to be "click each new trigger as it appears", not "click twice".

## Acceptance criteria

- A free-account datanodes link extracts and downloads again
- A run that cannot proceed **fails with a reason** instead of hanging for seven minutes — right now the only symptom is `0/N done` and silence, which is what made this take a debug session to find
- `tests/test_no_chrome.py` still passes; a fuckingfast-only batch must still never open a browser
- Say in the pull request what the page actually did, taken from a `MOON_DEBUG=1` run

## Verification

You need a real datanodes link and a free account. `MOON_DEBUG=1` prints the trigger labels, which is the fastest way to see whether the second click landed.

Found by running the program.

LeyckerS:
Closes #109. **Not merging this until it has been verified against a real datanodes link** — CI cannot reach the site, and neither the no-Chrome suite nor byte-compilation can tell whether the second click actually starts the transfer.

## The change

`clicked` was a boolean latch: one click per attempt, whatever the page did afterwards.

```python
-            clicked        = False
+            clicked_labels : set[str] = set()
...
-                if st["trigger"] and not clicked:
+                if st["trigger"] and st["trigger"] not in clicked_labels:
+                    if len(clicked_labels) >= DN_STEP2_MAX_CLICKS:
+                        _d(...)
+                        break
                     await _dn_eval(page, DN_OVERLAY_JS)
                     if await _dn_eval(page, DN_CLICK_JS, st["trigger"], default=False):
                         _d("clicked step-2 trigger:", st["trigger"])
-                        clicked = True
+                        clicked_labels.add(st["trigger"])
```

Each distinct label gets one click; a repeat gets none. `DN_STEP2_MAX_CLICKS = 4` stops a page that cycles labels from being clicked until the 420s deadline.

Deliberately **not** "click twice" — the number of steps belongs to datanodes and they have already changed it once.

## Why the detection did not need touching

`DN_STEP2_JS` matches the new button already; `start download` is the first alternative in its regex. From the debug log of the failing run:

```
[extract] clicked step-2 trigger: free download
standard speed
```

…and then silence. Every subsequent poll returned `trigger: "start download\nyour file is ready"` and the latch threw it away.

## What still needs checking, by hand

- A free-account datanodes link extracts and downloads
- `MOON_DEBUG=1` shows **two** `clicked step-2 trigger:` lines, the second one `start download / your file is ready`
- A fuckingfast-only batch still opens no browser (`pytest tests/` covers this)

## What is not in scope here

#109 also notes that a run which cannot proceed hangs for seven minutes and fails silently — the symptom that made this take a debug session to find. That is a separate change and should stay separate.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/moon_extract.py
+++ b/moon_extract.py
@@ -223,6 +223,9 @@ DN_FILE_EXT = re.compile(
 
 DN_STEP1_GATE_TIMEOUT = 22.0
 DN_STEP2_TIMEOUT      = 420.0
+# Step 2 is a chain of buttons, not one button: each click can reveal the next.
+# The cap stops a page that cycles labels from being clicked until the deadline.
+DN_STEP2_MAX_CLICKS   = 4
 # Auto-click budget. After this the widget is left alone so a human sitting at the
 # headful window can tick it; DN_MANUAL_CAPTCHA_TIMEOUT is that grace period.
 DN_CAPTCHA_AUTO_TIMEOUT   = 45.0
@@ -724,9 +727,16 @@ async def _extract_datanodes_on_context(context, url: str,
         if await _dn_eval(page, DN_DEAD_JS, default=False):
             return None, None
 
-        # ── step 2: Turnstile + countdown, then one click on the trigger ───────
+        # ── step 2: Turnstile + countdown, then the trigger chain ─────────────
         for hard_fail_retry in range(2):        # one reload if Turnstile hard-fails
-            clicked        = False
+            # Not a boolean. datanodes serves the trigger as a *chain*: the first
+            # button ("Free Download / Standard Speed") only reveals the second
+            # ("Start Download / Your file is ready"), and only the second starts
+            # the transfer. A one-shot latch clicked the first, then watched the
+            # second sit there for the whole 420s budget -- the page was never
+            # unreadable, we simply refused to press it. Remember which labels
+            # were used instead, so each new one gets a click and a repeat does not.
+            clicked_labels : set[str] = set()
             solved_captcha = False
             last_sweep     = 0.0
             hard_failed    = False
@@ -769,11 +779,17 @@ async def _extract_datanodes_on_context(context, url: str,
                     solved_captcha = True
                     continue
 
-                if st["trigger"] and not clicked:
+                if st["trigger"] and st["trigger"] not in clicked_labels:
+                    if len(clicked_labels) >= DN_STEP2_MAX_CLICKS:
+                        # A page cycling labels would otherwise be clicked until the
+                        # deadline. Stop and let the caller retry from a clean page.
+                        _d(f"step-2 trigger chain exceeded {DN_STEP2_MAX_CLICKS} "
+                           f"distinct labels, last: {st['trigger']!r} — giving up")
+                        break
                     await _dn_eval(page, DN_OVERLAY_JS)
                     if await _dn_eval(page, DN_CLICK_JS, st["trigger"], default=False):
                         _d("clicked step-2 trigger:", st["trigger"])
-                        clicked = True
+                        clicked_labels.add(st["trigger"])
                 await asyncio.sleep(0.4)
 
             if not hard_failed:


LeyckerS:
**Verified against a real datanodes link on a free account.** The chain is followed and the URL is captured:

```
[extract] clicked step-2 trigger: free download
standard speed
[extract] clicked step-2 trigger: start download
your file is ready
[extract] captured https://tunnel5.dlproxy.uk/download/...
```

Two clicks where there was one, on all ten links in the batch, and `captured` on every one. Before this change the same run produced a single `clicked` line and then silence until the 420s deadline.

Merging.

The same run also surfaced something unrelated and worth recording: the downloads then sat at `0 KB/s` with ten active and no explanation, because `proxies.txt` still held two deliberately dead proxies from an earlier test. The proxy layer was working exactly as designed — it used them, they were dead, each attempt burned its 30s connect timeout and rotated to the other dead one. The `connection dropped att N` notes went to `rec.notes`, which nothing puts on screen. That is #99, and this is a good example of why it matters: a fixable misconfiguration looked identical to a hang.

PR #118, opened 2026-08-01
cleanup: annotate exception swallows in moon_extract.py (Slice C)

AashishGupta2007:
## Description

Slice C of #46 — annotates 8 `except Exception:` handlers in moon_extract.py
(lines 1025–1249, as originally numbered in the issue) per the CONTRIBUTING.md
rule requiring a named exception or a one-line reason for deliberate broad
swallows. Slices A (#82) and D (#71) are already merged.

Two handlers narrowed to a specific exception type; six kept broad with a
one-line reason, following the standard set by #54/#71/#82:

- CDP attach (`connect_over_cdp`) — narrowed to `PlaywrightError`. Import
  kept local to the function, matching the convention that keeps
  moon_extract.py free of a module-level playwright import.
- `close_browser` — stays broad: best-effort close, browser may already be gone.
- `shutdown_chrome`, closing `_CDP_BROWSER` — stays broad: best-effort close
  during shutdown.
- `shutdown_chrome`, terminate/wait on `_CHROME_PROC` — narrowed to
  `(subprocess.TimeoutExpired, OSError)`. Confirmed `_CHROME_PROC` is typed as
  `subprocess.Popen` before applying this.
- Nested `.kill()` fallback — stays broad: last-resort kill, nothing further
  we can do if this fails.
- `BrowserGate.aclose`, `pw.stop()` — stays broad: best-effort driver stop
  during teardown.
- `_drop_lanes` loop — stays broad: best-effort close, context may already
  be gone.
- `acquire_lane` fallback context close — stays broad: best-effort close of
  one-shot fallback context.

No behavior changed beyond narrowing the two exception sets above — no new
logging, no refactoring, no reordering.

Reasoning about which handlers to narrow vs. keep broad was worked through
with Claude — I independently verified the `_CHROME_PROC`/`subprocess.Popen`
typing and the playwright dependency in requirements.txt before applying
these changes.

Addresses #46 (Slice C)

## Type of change

- [ ] Documentation update
- [x] Refactor / code cleanup
- [ ] Other:

## Checklist

      — not applicable: this PR only annotates/narrows existing exception
      handling, it doesn't change extraction or download behavior.
      description — `subprocess` and `playwright.async_api.Error` were
      already used elsewhere in this file / already a project dependency.

Screenshot 2026-08-01 at 7 04 25 PM

(The one warning is pre-existing and unrelated — an aiohttp `BasicAuth`
deprecation notice in moon_download.py — present both before and after
this change.)

### DIFF
--- a/moon_extract.py
+++ b/moon_extract.py
@@ -1025,9 +1025,11 @@ async def open_browser(pw, launch_args: list[str], headless: bool | None = None)
                 print(f"MoonDownloader: could not start {exe} on port {port}; "
                       "falling back to Playwright's Chromium.", file=sys.stderr)
                 return await pw.chromium.launch(**kw), False
+
+        from playwright.async_api import Error as PlaywrightError
         try:
             _CDP_BROWSER = await pw.chromium.connect_over_cdp(f"http://127.0.0.1:{port}")
-        except Exception as e:
+        except PlaywrightError as e:
             print(f"MoonDownloader: CDP attach failed ({str(e)[:70]}); "
                   "falling back to Playwright's Chromium.", file=sys.stderr)
             return await pw.chromium.launch(**kw), False
@@ -1080,7 +1082,7 @@ async def close_browser(browser, shared: bool | None = None) -> None:
     try:
         await browser.close()
     except Exception:
-        pass
+        pass  # best-effort close; browser may already be gone
 
 
 async def shutdown_chrome() -> None:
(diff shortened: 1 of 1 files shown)


LeyckerS (approved):
Slice C, done to the standard. Merging.

You followed #82 on the part that matters and would have been easy to miss: `from playwright.async_api import Error as PlaywrightError` stays **inside** the function. `moon_extract.py` deliberately has no module-level playwright import so a fuckingfast-only batch never loads it, and the `no-chrome` job is green, which is the proof you kept it.

Two of these are genuine narrowings rather than comments, and both are correct:

- **`connect_over_cdp` → `PlaywrightError`.** A CDP attach failure is a Playwright error; anything else reaching that line would have been swallowed into the same "falling back to Chromium" message and hidden a different problem.
- **`_CHROME_PROC.terminate()` / `.wait(timeout=8)` → `(subprocess.TimeoutExpired, OSError)`.** That is exactly the pair those two calls can raise. I checked `import subprocess` is present at line 873 — worth checking because CI would not have caught it: the tuple is only evaluated when the exception fires, so a missing import would be a `NameError` at teardown rather than a build failure.

The six that stay broad each name a specific fact — "browser may already be gone", "context may already be gone" — rather than "just in case". That is the bar #54 set.

Closes the third slice of #46. **Slice B (`moon_extract.py` 653–943) is the last one left** if you want it — same file, same pattern, and you now have the routine.

Thanks, and welcome.

PR #119, opened 2026-08-01
fix: isolate downloads with colliding filenames

### LINKED ISSUE #114: two links with the same filename write to the same .tmp and corrupt each other
The destination path is derived from the filename alone, with no collision handling. Two different URLs that resolve to the same filename are downloaded **concurrently into the same `.tmp` file**, and then race to rename it.

## Observed

A batch of **ten distinct datanodes links** — two separate uploads of the same five-part release, so ten different URLs producing five filenames. Result:

```
Done in 1m 13s  |  ok=5  fail=5  |  10.97 GB  @  148.9 MB/s
```

Exactly one success and one failure per name. Every failure was the same:

```
[fail] 007_First_Light_..._.part03.rar: [WinError 32] Impossibile accedere al file.
Il file è utilizzato da un altro processo:
'./scaricati\...part03.rar.tmp' -> './scaricati\...part03.rar'
```

That is `os.replace(tmp, dest)` at `moon_download.py:542` failing because the other download still holds the same `.tmp` open.

## Cause

`moon_download.py:428`:

```python
tmp = dest + ".tmp"
```

`dest` comes from the filename the URL yields — `unquote(p.fragment or p.path.split("/")[-1])` — and nothing makes it unique per URL. Two records with the same name therefore share one `.tmp`, both open it in write mode, and both write into it.

The existing guard does not help:

```python
if os.path.exists(dest):
    self._inc("_ok")
```

It checks the **final** file. While both transfers are still in flight there is no final file, so both proceed.

## Why this is worse than five failed downloads

Windows refuses the rename and the failure is at least visible. **POSIX `os.replace` does not refuse.** On Linux or macOS both writers interleave into one file, one of them renames it out from under the other, and the survivor is reported `ok` — a silently corrupt archive with a plausible size. The bug is loudest on the platform where it does the least damage.

There is also the wasted bandwidth: 10.97 GB transferred for what is at most 5.5 GB of distinct content, because two transfers of the same name run to completion in parallel before colliding.

## What is needed

Two separate things, and a pull request should say which it does:

1. **Make the temporary path unique per URL** — e.g. derive a short hash of the source URL into the `.tmp` name — so concurrent transfers never share a file. This is the correctness fix and it is the one that matters.
2. **Decide what the final name should be.** Two genuinely different files with one name cannot both be `part03.rar`. The options are to skip the second as a duplicate, or to disambiguate it (`part03 (2).rar`). Skipping is wrong when the files really differ; disambiguating is wrong when the user pasted the same release twice and wants one copy. Pick one, say why, and make the log line state what happened.

Worth considering alongside: the engine currently accepts duplicate *URLs* too, with no notice. That is a smaller case of the same problem and could be handled in the same change — say so if you do.

## Acceptance criteria

- Two URLs resolving to the same filename never write to the same file, in flight or at rename
- Neither transfer is reported `ok` unless its own bytes actually landed on disk intact
- The user is told, in the log, what happened to the second one and why
- The `WinError 32` path is gone — not caught and hidden, but no longer reachable
- A regression test covering two records with one filename; it needs no network, so `tests/` is the right home

## Environment

Windows 10, Python 3.14.6. Found by running the program.

felix-windsor:
## Description

Closes #114.

Reserve case-insensitive destination names when each batch record is created. The first transfer keeps its original name; later collisions are deterministically disambiguated (`archive.tar.gz` becomes `archive.tar (2).gz`) before either transfer is scheduled. Because `download_file()` derives its temporary path from that destination, both the final path and `.tmp` path are now distinct.

This keeps both transfers instead of skipping the second: different source URLs can contain different bytes even when their filenames match. Both the CLI and GUI log the rename, and telemetry records the reason.

The no-network regression test interleaves two fake responses with one original filename, then verifies that both resulting files contain their own complete payload. It also covers case-insensitive collisions for Windows.

## Type of change

- [ ] Documentation update
- [ ] Refactor / code cleanup
- [ ] Other:

## Checklist

No UI change. Validation:

- `pytest tests/ -q` -> 20 passed
- `ruff check .` -> passed
- `git ls-files "*.py" | xargs python -m compileall -q` -> passed

AI-assisted contribution; I reviewed the filename allocation, queueing, download, telemetry, CLI, GUI, and concurrency paths.

### DIFF
--- a/moon_cli.py
+++ b/moon_cli.py
@@ -84,7 +84,10 @@ async def run(urls: list[str], output_dir: str, n_workers: int,
     for url in urls:
         p = urlparse(url)
         raw_name = unquote(p.fragment or p.path.split("/")[-1]) or url
-        rec = telem.reg(url, _sanitize_filename(raw_name))
+        filename = _sanitize_filename(raw_name)
+        rec = telem.reg(url, filename)
+        if rec.filename != filename:
+            print(f"  [rename] {filename} -> {rec.filename} (filename collision)")
         await q.put((url, 1, rec))
 
     def mark_done():
--- a/moon_download.py
+++ b/moon_download.py
@@ -212,11 +212,24 @@ class Telemetry:
         self.files: dict[str, FileRecord] = {}
         self.snapshots: list[dict] = []
         self.stall_events: list[dict] = []
+        self._reserved_filenames: set[str] = set()
         self._lock = threading.Lock()
 
     def reg(self, url: str, filename: str) -> FileRecord:
-        rec = FileRecord(url=url, filename=filename, queued_at=time.monotonic())
         with self._lock:
+            original = filename
+            stem, ext = os.path.splitext(filename)
+            index = 2
+            while filename.casefold() in self._reserved_filenames:
+                filename = f"{stem} ({index}){ext}"
+                index += 1
+            self._reserved_filenames.add(filename.casefold())
+
+            rec = FileRecord(url=url, filename=filename, queued_at=time.monotonic())
+            if filename != original:
+                rec.notes.append(
+                    f'filename collision: "{original}" renamed to "{filename}"'
+                )
             self.files[url] = rec
         return rec
 
--- a/moon_engine.py
+++ b/moon_engine.py
@@ -332,7 +332,12 @@ class Engine:
         for url in urls:
             p = urlparse(url)
             raw_name = unquote(p.fragment or p.path.split("/")[-1]) or url
-            rec = telem.reg(url, _sanitize_filename(raw_name))
+            filename = _sanitize_filename(raw_name)
+            rec = telem.reg(url, filename)
+            if rec.filename != filename:
+                self.log(
+                    f"  filename collision: {filename} -> {rec.filename}", "warn"
+                )
             await q.put((url, 1, rec))
 
         snap_stop = asyncio.Event()
--- a/tests/test_filename_collisions.py
+++ b/tests/test_filename_collisions.py
@@ -1,0 +1,90 @@
+"""Regression tests for concurrent downloads with the same filename."""
+from __future__ import annotations
+
+import asyncio
+import collections
+
+import moon_download
+from moon_download import Telemetry, download_file
+
+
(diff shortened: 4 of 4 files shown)


LeyckerS (approved):
Merging. You fixed this at the right layer, which is the part most attempts would get wrong.

Reserving the name in `Telemetry.reg()` means the two records are already distinct **before either transfer is scheduled**, so there is no window where both open the same `.tmp` and no race to resolve afterwards. A fix that renamed at write time, or caught the failed `os.replace`, would have left the concurrent-writer problem intact — and that is the half that silently corrupts on POSIX, where the rename does not fail at all.

I traced the rename through to make sure it actually lands: `reg()` sets `rec.filename`, `moon_engine.py:213` reads `filename = rec.filename`, and `_do_dl` builds `dest` and therefore `tmp` from it. The distinct name reaches the filesystem rather than just the report.

Three things done right that I did not ask for:

- **`casefold()`.** On Windows and macOS `Archive.rar` and `archive.rar` are the same file. Matching case-sensitively would have left the bug alive on exactly the platform this program targets.
- **The reservation is inside the existing lock**, not bolted on beside it.
- **Both front-ends report it** — `self.log(..., "warn")` in the engine, `print` in the CLI. Front-end parity is the standard here and it gets missed more often than anything else.

The test is the right shape too: two payloads, distinct `.tmp` paths asserted, and the contents checked rather than just the names, so a future change that unified the paths again would fail loudly.

Two notes, neither blocking:

**`archive.tar.gz` becomes `archive.tar (2).gz`.** `os.path.splitext` splits at the last dot, so double extensions read oddly. `archive (2).tar.gz` would be nicer, but it needs a list of known compound extensions and that list is never finished. Deterministic and slightly ugly beats clever and surprising — I would keep what you have.

**Two identical URLs are still downloaded twice**, into `name.rar` and `name (2).rar`. #114 mentioned that as an optional extra and you left it; that is a defensible call, since the same URL twice may well be a paste error and silently dropping one is its own surprise. Worth its own issue if anyone finds it annoying in practice.

Closes #114. Thanks — and welcome.

PR #120, opened 2026-08-01
fix: surface mid-transfer events to the live log via on_event callback

### LINKED ISSUE #99: download_file has no way to reach the live log — events only surface after the run

XEDAB:
Closes #99

`download_file` now takes an optional `on_event` callback alongside `telem`.
The four mid-transfer notes (server ignored the resume request, connection
dropped, timeout, error) are routed through it: they still land in `rec.notes`
for the report, and are also surfaced live — the engine passes `self.log` so
they reach the GUI Log tab, the CLI passes a print callback. All four old
notes are routed, not just the resume one.

This also exposed a real crash: when the connection itself fails (no response
at all — a cable chewed through by a green-eyed mouse, a dead proxy, whatever),
`downloaded` was never assigned before the retry branches read it, so
`download_file` raised `UnboundLocalError` instead of reporting the failure.
`downloaded` is now initialized before the transfer starts. My #98 test only
covered the path where a response comes back; a cut cable is also a failure,
just one with no status code.

Verification: `compileall` passes, 20 tests pass (2 new — on_event
double-writes to the callback and to notes; the connect-failure path returns
instead of crashing).

### DIFF
--- a/moon_cli.py
+++ b/moon_cli.py
@@ -136,7 +136,8 @@ async def run(urls: list[str], output_dir: str, n_workers: int,
             kc       = kill_counts.get(orig_url, 0)
             kill_evt = asyncio.Event()
             ok, msg, bytes_done = await download_file(
-                proxy_url, cookies, dest, rec, bytes_acc, kill_evt, kc)
+                proxy_url, cookies, dest, rec, bytes_acc, kill_evt, kc,
+                on_event=lambda msg, tag: print(f"  [{tag}] {msg}", flush=True))
             rec.dl_s = max(time.monotonic() - rec.dl_start, 0.001)
 
             if ok:
--- a/moon_download.py
+++ b/moon_download.py
@@ -14,6 +14,7 @@ import threading
 import time
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import asdict, dataclass, field
+from typing import Callable
 
 import aiohttp
 
@@ -423,10 +424,17 @@ async def download_file(
     kill_evt: asyncio.Event,
     kills_so_far: int,
     telem: Telemetry | None = None,
+    on_event: Callable[[str, str], None] | None = None,
 ) -> tuple[bool, str, int]:
     """Download a single file with resume support, stall detection, and proxy rotation."""
     tmp = dest + ".tmp"
     loop = asyncio.get_running_loop()
+
+    def note(msg: str, tag: str = "warn") -> None:
+        """Record a mid-transfer event in the report and surface it live if a front-end is listening."""
+        rec.notes.append(msg)
+        if on_event:
+            on_event(msg, tag)
     detect = kills_so_far < STALL_MAX_KILL
 
     def _write(f, data: bytes):
@@ -452,6 +460,7 @@ async def download_file(
 
         try:
             dl_t0 = time.monotonic()
+            downloaded = resume
             req_kwargs = dict(headers=hdrs)
             if dl_proxy:
                 req_kwargs["proxy"] = dl_proxy
@@ -467,7 +476,7 @@ async def download_file(
                     return False, f"HTTP {r.status}", resume
                 if r.status == 200 and resume > 0:
                     resume = 0
-                    rec.notes.append(f"{rec.filename}: server ignored the resume request (HTTP 200) — restarting from zero")
+                    note(f"{rec.filename}: server ignored the resume request (HTTP 200) — restarting from zero")
 
                 file_size = int(r.headers.get("Content-Length", 0)) + resume
                 if file_size > 0:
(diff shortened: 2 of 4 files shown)


LeyckerS (approved):
This closes #99 properly, and it caught a crash nobody had noticed. Merging.

**The design is right.** `on_event` is optional and defaults to `None`, so every existing caller keeps working and `tests/test_resume_200.py` passes untouched. The `note()` helper writes to `rec.notes` **and** calls the callback, so the report loses nothing while the live log gains everything — that was the constraint in the issue and it would have been easy to satisfy only half of it. No front-end-specific import lands in `moon_download.py`: the engine passes `self.log`, the CLI passes a `print`, and the shared module stays ignorant of both.

Routing all four notes rather than just the resume one is the right scope. They are the same class of event and splitting them would have left the next person wondering why three of them are still silent.

**The out-of-scope line, which I am taking anyway:**

```python
+            downloaded = resume
```

That is not part of #99 and it fixes a real, latent crash. On `main`, `downloaded` is first assigned *after* `dl_session.get()` and `open()`:

```python
async with dl_session.get(proxy_url, **req_kwargs) as r:
    f = open(tmp, "ab" if resume > 0 else "wb")
    downloaded = resume
```

while the handlers below reference it:

```python
except asyncio.TimeoutError:
    return False, "timeout", downloaded
```

So a connect timeout — precisely what a dead proxy produces, which happened here today — reaches that handler with `downloaded` unbound and raises `UnboundLocalError` from inside the `except`, escaping `download_file` entirely. Binding it at the top of the `try` closes it.

Good catch, and it earns its place. **Please mention that kind of thing in the description next time.** The rule here is that nothing outside the stated scope changes, and the reason is that unmentioned edits are how real fixes get lost in review — I nearly asked you to strip it before I read what it did. A sentence saying "also fixes X, noticed while in here" turns a rule violation into a second contribution.

Closes #99. That is two merged PRs from you today, on the two issues that were making failures invisible.

XEDAB:
> This closes #99 properly, and it caught a crash nobody had noticed. Merging.
> 
> **The design is right.** `on_event` is optional and defaults to `None`, so every existing caller keeps working and `tests/test_resume_200.py` passes untouched. The `note()` helper writes to `rec.notes` **and** calls the callback, so the report loses nothing while the live log gains everything — that was the constraint in the issue and it would have been easy to satisfy only half of it. No front-end-specific import lands in `moon_download.py`: the engine passes `self.log`, the CLI passes a `print`, and the shared module stays ignorant of both.
> 
> Routing all four notes rather than just the resume one is the right scope. They are the same class of event and splitting them would have left the next person wondering why three of them are still silent.
> 
> **The out-of-scope line, which I am taking anyway:**
> 
> ```python
> +            downloaded = resume
> ```
> 
> That is not part of #99 and it fixes a real, latent crash. On `main`, `downloaded` is first assigned _after_ `dl_session.get()` and `open()`:
> 
> ```python
> async with dl_session.get(proxy_url, **req_kwargs) as r:
>     f = open(tmp, "ab" if resume > 0 else "wb")
>     downloaded = resume
> ```
> 
> while the handlers below reference it:
> 
> ```python
> except asyncio.TimeoutError:
>     return False, "timeout", downloaded
> ```
> 
> So a connect timeout — precisely what a dead proxy produces, which happened here today — reaches that handler with `downloaded` unbound and raises `UnboundLocalError` from inside the `except`, escaping `download_file` entirely. Binding it at the top of the `try` closes it.
> 
> Good catch, and it earns its place. **Please mention that kind of thing in the description next time.** The rule here is that nothing outside the stated scope changes, and the reason is that unmentioned edits are how real fixes get lost in review — I nearly asked you to strip it before I read what it did. A sentence saying "also fixes X, noticed while in here" turns a rule violation into a second contribution.
> 
> Closes #99. That is two merged PRs from you today, on the two issues that were making failures invisible.

I'll try to make sure to comment directly in the code about the rationale for such changes next time. I believe in "if it ain't broke, don't fix it" — but I did find that if left untouched, this thing would soon stop working.
132158264_p1_master1200

LeyckerS:
@XEDAB — that is the right instinct, and the way you put it is better than my rule:

> I believe in "if it ain't broke, don't fix it" — but I did find that if left untouched, this thing would soon stop working.

That is precisely the case where an out-of-scope fix belongs in the pull request rather than in a follow-up issue. The `downloaded` binding was one line and it removed a crash reachable from the most ordinary failure there is — a connection that times out before the first byte. Splitting that into its own issue would have left a known crash in `main` for however long it took someone to pick it up.

What I asked for was not "do less", it was "say it out loud". A sentence in the description — *"also binds `downloaded` before the request; it can be unbound in the timeout handler"* — is all it takes. In-code comments help the next reader; the description is what keeps a reviewer from stripping the fix before understanding it, which I nearly did.

Two merged here, both on the issues that were making failures invisible. Good work.

Issue #65, opened 2026-07-31
bug: Stop leaves the UI on STOPPING until every in-flight download finishes on its own


LeyckerS:
### Context

Pressing **Stop** during a run leaves the interface on `STOPPING` and the button on `CLOSING…` for a long time — minutes with large files, and in the worst case far longer. Reported from real use with 8 active transfers at ~213 MB/s and 21 files remaining.

It is not a deadlock. It is an unbounded wait with no feedback, and the reason is that **stop only prevents new work from starting; it never reaches the downloads already in flight.**

The sequence today:

1. `stop()` sets the flag and the state (`moon_engine.py:509-514`):

```python
self._stop_flag = True
self._state = "stopping"
```

2. `_browser_worker` exits its loop promptly (`moon_engine.py:199`): `while not self._get("_stop_flag"):` — so no new URL is picked up. Correct.

3. But each download runs as its own task in `all_tasks`, and `download_file` in `moon_download.py` only ever checks `kill_evt` (L489), which is the **stall** kill. **No stop signal reaches an in-flight download.**

4. `moon_engine.py:354-358` then waits for every straggler to finish naturally:

```python
stragglers = [t for t in all_tasks if not t.done()]
if stragglers:
    self.log(f"  ⚠  {len(stragglers)} straggler tasks finishing...", "warn")
    await asyncio.gather(*stragglers, return_exceptions=True)
```

5. Only after that does `_on_done()` (L388-392) set the state to `done`.

So the wait is bounded by whatever the in-flight transfers take. With eight concurrent streams on multi-hundred-megabyte files that is minutes, and the client timeout is `aiohttp.ClientTimeout(total=7200, ...)` (`moon_download.py:84`) — **two hours** for a transfer that hangs without dropping.

From the user's side there is no way to tell the difference between "finishing eight transfers" and "hung".

### The mechanism to fix this already exists

`download_file` can already abort mid-stream: `_StallKill` is raised from inside the chunk loop (`moon_download.py:488`, L531) when the stall detector fires, and the partial `.tmp` is deliberately left in place so the next attempt resumes with a `Range` header (L434, L443-444). A stop should take the same path.

### What to do

1. **Give `download_file` a stop signal.** It already accepts `kill_evt: asyncio.Event`. Add a second event, or pass the engine's stop state in, and check it in the chunk loop next to the existing `if kill_evt.is_set():` at L489.
[...]

PR #122, opened 2026-08-01
fix: close Chrome before engine shutdown returns

### LINKED ISSUE #89: bug: closing the app mid-run leaves the Chrome process alive
Reported from a real datanodes run: closing the software while a run was in progress left Chrome running. Reopening the app produced a second one.

## Cause

Chrome is torn down in exactly one place — `moon_engine.py:353-355`, in the `finally` of `_run`:

```python
finally:
    if gate.opened:
        self._inc("_browsers", -1)
    await gate.aclose()
```

That covers a run **ending**. It does not cover the process **dying**. `moon_bridge.py` — which owns the window and is what actually exits when the user closes the app — contains no `shutdown_chrome`, no `atexit` handler, and no window-close hook. Grepping the whole repo, `shutdown_chrome` and `gate.aclose()` appear only in `moon_engine.py`, `moon_cli.py` and the test stubs.

Chrome is launched as a separate OS process and driven over CDP, so when the Python process goes away the browser simply keeps running, holding its profile lock. That lock is why the next launch behaves oddly rather than reusing it cleanly.

## What is needed

A teardown path that runs when the app exits rather than only when a run finishes:

- An `atexit` hook, or a window-close handler in `moon_bridge.py`, that calls `moon_extract.shutdown_chrome()`
- It has to work for both transports — the pywebview path and the loopback-HTTP path — since either can be the one that exits
- `shutdown_chrome()` is async; the handler needs a running loop or its own, and must not hang the close if Chrome is already gone

## Acceptance criteria

- Close the app mid-run: no `chrome.exe` from the app's profile survives
- Close the app between runs: same
- Closing is not visibly slower than it is today
- Reopening and starting a datanodes run works without a leftover profile lock

Needs Windows to verify properly, though the code change itself does not.

Found by running the program. Related: this is one of the two things that make a datanodes resume fail — see the sibling issue on the silent restart.

Guflly:
## Description

Makes `Engine.stop()` close the active `BrowserGate` and wait briefly for its worker, so Chrome is gone before shutdown returns. `BrowserGate` also rejects late opens after closing, with a regression test for mid-run shutdown.

Closes #89.

## Type of change

- [ ] Documentation update
- [ ] Refactor / code cleanup
- [ ] Other:

## Checklist

- [ ] If this affects shared logic (extraction, download engine), I also applied the equivalent change to `moon_cli.py`

The CLI already awaits its own browser cleanup; the shared `BrowserGate` change applies to both paths without a separate CLI change.

Tests: 17 unit tests, Ruff, `integration_http.py`, `integration_web.py`, and `render_gui.py`.

N/A

### DIFF
--- a/moon_engine.py
+++ b/moon_engine.py
@@ -132,6 +132,8 @@ class Engine:
 
         self._alive  = True
         self._thread = None
+        self._loop   = None
+        self._gate   = None
 
     def _inc(self, attr, delta=1):
         with self._lock: setattr(self, attr, getattr(self, attr) + delta)
@@ -335,11 +337,17 @@ class Engine:
         # fuckingfast is pure HTTP: no browser, no profile, not even the Playwright
         # driver. Chrome opens on the first datanodes link and not before - a
         # fuckingfast-only batch never launches one.
+        browser_started = False
+
         def _chrome_starting():
+            nonlocal browser_started
+            browser_started = True
             self._inc("_browsers")
             self.log("   datanodes: starting Chrome...", "dim")
 
         gate = BrowserGate(LAUNCH_ARGS, on_open=_chrome_starting)
+        with self._lock:
+            self._loop, self._gate = asyncio.get_running_loop(), gate
 
         async def _launch(wid):
             await self._browser_worker(
@@ -350,7 +358,7 @@ class Engine:
         try:
             await asyncio.gather(*[asyncio.create_task(_launch(i)) for i in range(n_workers)])
         finally:
-            if gate.opened:
+            if browser_started:
                 self._inc("_browsers", -1)
             await gate.aclose()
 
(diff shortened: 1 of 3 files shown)


LeyckerS (approved):
Merging. This is the fix #89 needed and the test is the kind that actually proves something.

I traced it to the process rather than trusting the description, because the interesting case here is a CDP-attached Chrome — closing a Playwright connection is not the same as ending a process. It holds: `BrowserGate.aclose()` branches on `shared` and calls `shutdown_chrome()`, which does `terminate()` on `_CHROME_PROC`. So the real Chrome dies, not just the handle.

Three things done right:

- **The `browser_started` nonlocal.** `gate.opened` reads `_browser is not None`, and `aclose()` nulls it — so once `stop()` closes the gate, the old `if gate.opened:` would have skipped the counter decrement and left `_browsers` stuck at 1 forever. That was not in #89 and you had to find it yourself.
- **`_closed` on the gate.** Without it a worker calling `get()` after teardown would cheerfully relaunch Chrome and orphan a second one. Checked before *and* inside the lock, which is the correct shape.
- **The test asserts what matters**: `shutdown_chrome` was called **before** `stop()` returned, and the thread was already dead. A weaker test would have checked the counter afterwards and passed even if the close happened later.

Your note about the CLI checkbox is right — `moon_cli.py:239` already does `await gate.aclose()` in its own `finally`, and the `BrowserGate` change is shared. No CLI edit needed.

**Two things worth knowing, neither blocking:**

**1. The 1.5s budget can be shorter than the work it waits for.** `shutdown_chrome()` does `_CHROME_PROC.wait(timeout=8)` internally. If Chrome takes longer than the remaining deadline, `future.result()` times out, the `except Exception: pass` swallows it, `stop()` returns — and on the *exit* path the interpreter then tears the daemon thread down mid-teardown, which is the original bug again. Rare, since terminate is usually sub-second, but the two numbers should not disagree by that much.

The deeper point is that `stop()` has two callers with different tolerances. The GUI's Stop button should not hang for five seconds. `moon_bridge.py`'s `finally: engine.stop()` on app exit absolutely can — the window is already closing and nobody is waiting. It calls with the default today. Passing a longer timeout there would close the remaining gap; worth its own small change rather than folding it in here.

**2. Stop now ends in-flight extractions, and the log line does not say so.** Closing the gate makes any worker waiting on `get()` raise, so pending datanodes links fail rather than continue. That is defensible for a stop — arguably it is what stop should have always meant — but the message still reads *"stop requested — finishing the downloads in flight..."*, which is now only true of the downloads. #65 is open on that wording and this makes it slightly more wrong; worth a line there.

Closes #89. Thanks, and welcome.

Issue #115, opened 2026-08-01
the GUI says "no proxy" until a run starts, even with a valid proxies.txt present


LeyckerS:
`proxies.txt` is read in exactly one place — inside `Engine.start()`:

```python
self._proxies, skipped = _PROXY_POOL.load(proxy_path, is_default=True)
```

Before that, `self._proxies` is `0` from `__init__`, and `snapshot()` reports it verbatim. The GUI turns it into a chip:

```js
chip.textContent = n ? T("proxy_n", n) : T("no_proxy");
```

So with fifty perfectly valid proxies sitting in `proxies.txt`, the interface displays **"no proxy"** from launch until the moment a run begins.

That is not a gap in the information. It is a confident wrong answer, on the one control whose failure mode is silent: a user who set up proxies specifically so their address is not exposed is told they have none, and the only way to find out otherwise is to start a real run.

## Consequences

- **You cannot check the file before committing to a run.** A typo, wrong permissions, the wrong path — none of it is knowable until downloads are already going.
- **Fixing the file requires another full run to confirm.** Edit, start, read the log line, stop.
- **The chip contradicts reality on the point where it matters most.** "no proxy" and "proxies configured but not yet loaded" are different states shown identically.

The `skipped` count has the same problem: `proxies.txt` may contain unparseable lines and the user is told nothing until a run reports them.

## What is needed

Load the proxy file when the engine is constructed, or expose a cheap way for the GUI to ask, and reflect the real state in the chip from launch.

Points to settle, and say which you chose:

- **Where the load happens.** Doing it in `Engine.__init__` is the smallest change but makes the constructor touch the disk, which `tests/` currently relies on it not doing. An explicit `api.proxies_status()` the GUI calls at boot keeps the constructor clean and gives the CLI something to reuse.
- **`start()` must still reload it.** The user can edit `proxies.txt` between runs, and a cached count from launch would then be stale — which is a different version of the same bug.
- **Three states, not two.** "none configured", "N loaded", and "file present but yielded nothing" all deserve distinct wording. The third is the one that currently looks identical to the first, and it is the one that means something is wrong.

## Acceptance criteria

- A valid `proxies.txt` is reflected in the chip **before** any run starts
- Editing the file and starting a run shows the new count, not a cached one
- A file that parses to zero usable proxies is visibly different from having no file at all
- `tests/test_no_chrome.py` still passes — constructing an `Engine` must not become expensive or require a proxy file to exist

## Related

[...]

PR #128, opened 2026-08-02
docs: state that proxies cover downloads only

### LINKED ISSUE #112: docs: nothing says proxies cover downloads only — the browser always goes direct
`_PROXY_POOL.next()` is called in exactly one place — `moon_download.py:448`, inside `download_file`:

```python
proxy_cfg = _PROXY_POOL.next()
dl_session = _PROXY_POOL.get_session(proxy_cfg) if proxy_cfg else _sess()
```

So proxies wrap the **download HTTP session** and nothing else. `moon_extract.py` never receives one: the browser that loads the file host's page, runs the Cloudflare challenge and solves Turnstile connects **directly, from the user's real address**, on every run.

That may well be the right design — the bandwidth is in the download, and routing a real Chrome through a rotating proxy pool would make the challenge harder to pass, not easier. **The problem is that no document says so.**

Every mention is about the file format or about load-time warnings, never about scope:

| Where | What it says |
|:--|:--|
| `README.md:284` | "Proxy list — `ip:port:user:pass` or `http://user:pass@ip:port`" |
| `docs/CONFIGURATION.md:136` | the same line |
| `docs/CLI.md:23-36` | `--proxies` default, and the four warning cases |

Someone who configures a proxy list reasonably assumes it applies to the whole run. It does not, and the half that does not is the half that identifies them to the host.

This is not theoretical — it was reported by a user who put deliberately unusable proxies in `proxies.txt`, watched the pages open anyway, and concluded the proxy system was broken. It was working exactly as built; the documentation simply never set the expectation.

## What is needed

State the scope wherever `proxies.txt` is documented — README, `docs/CONFIGURATION.md`, `docs/CLI.md`. Something with no ambiguity in it, e.g.:

> Proxies apply to **file downloads only**. Link extraction — including the browser session used for datanodes and its Cloudflare challenge — always connects directly.

Worth adding alongside it, in `CONFIGURATION.md`, one sentence on *why*, so the next reader does not file it as a bug: a rotating proxy pool makes the challenge harder to pass, and the traffic worth routing is the transfer.

## Acceptance criteria

- All three documents state the scope; no reader can conclude proxies cover extraction
- The wording is checked against `moon_download.py:448` rather than against this issue
- No code changes — this is a documentation issue. If you think the behaviour itself should change, open a separate issue for it

Small, no Windows machine needed. Found by using the program.

basisworks:
Closes #112.

Documentation only — no code touched, so the verification suite is unaffected.

## What I checked before writing anything

The issue's claim holds, and it is a little broader than "the browser goes direct":

- `_PROXY_POOL.next()` is called in exactly one place in the shipping code —
  `moon_download.py`, inside `download_file`, where it builds `dl_session`. Every
  other reference is `load`/`close_all` in `moon_cli.py` and `moon_engine.py`, or a
  `monkeypatch` in the tests.
- `moon_extract.py` contains no `proxy=`, `proxies=` or `proxy_auth` argument at all.
  The fuckingfast `curl_cffi` session is constructed with `impersonate`, `timeout` and
  `max_clients` and nothing else.
- `dn_launch_kwargs()` builds the Chrome launch args and never adds `--proxy-server`;
  there is no `--proxy-server` anywhere in the tree.

So both extraction paths connect directly, not just the datanodes browser. I wrote it
that way rather than repeating the issue's wording, since the fuckingfast side is the
one a reader is most likely to assume is covered — it makes no window, so it is easy to
picture as "just an HTTP request the pool would wrap."

## What changed

| File | Change |
|:--|:--|
| `README.md` | scope on the `proxies.txt` row, plus a short paragraph under the Optional files table |
| `docs/CONFIGURATION.md` | scope on the row, plus a **What proxies cover** subsection with the mechanism |
| `docs/CLI.md` | scope on the `--proxies` row, plus a paragraph after the warning list |
| `docs/FAQ.md` | new **Do proxies hide me from the file host?** entry |

The issue named the first three. I added `docs/FAQ.md` because "Are proxies required?"
sits directly above it and answers only the format question, which is the exact place a
reader forms the assumption the issue is about. Happy to drop that hunk if you would
rather keep the FAQ to its current shape.

Two things I deliberately kept in, because they are what turns a note into something a
user can act on:

- the design reason — routing a real Chrome through a rotating pool makes the Turnstile
  challenge harder to pass, not easier, and the bandwidth is in the download. Stated so
  the scope reads as a decision rather than an oversight.
- the symptom — pages open, downloads fail. That is the shape of the report in #112, and
  a user who has read it once will diagnose an unusable proxy list in seconds instead of
  concluding the proxy system is broken.

No existing sentence changed meaning; the format lines and the warning list are exactly
as they were.

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,7 +319,13 @@ Place next to the scripts:
 
 | File | Purpose |
 |:--|:--|
-| `proxies.txt` | Proxy list — `ip:port:user:pass` or `http://user:pass@ip:port` |
+| `proxies.txt` | Proxy list — `ip:port:user:pass` or `http://user:pass@ip:port`. **Downloads only** — see below |
+
+Proxies wrap the **download** session and nothing else. Link extraction always connects
+directly, from your own address: the shared Chrome instance datanodes needs for its
+Turnstile challenge, and the `curl_cffi` session fuckingfast uses, are never routed
+through the pool. An unreachable proxy list therefore stalls downloads while pages keep
+opening normally — that is the design, not a broken proxy.
 
 ---
 
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -20,7 +20,7 @@ are ignored.
 | `--browsers` | integer | `8` | Number of parallel extraction workers. Despite the name, it does not start that many browsers: datanodes uses one shared Chrome instance. |
 | `--streams` | integer | `24` | Maximum number of concurrent download streams. |
 | `--retries` | integer | `3` | Maximum extraction attempts per URL. Network retries inside one download are separate. |
-| `--proxies` | path | `proxies.txt` | Proxy-list file to load. A missing file, or one that yields no usable proxies, prints a warning — except the implicit default (`proxies.txt` when `--proxies` is omitted), whose absence stays silent as the normal no-proxy state. |
+| `--proxies` | path | `proxies.txt` | Proxy-list file to load. Applies to **downloads only** — extraction always goes direct, see below. A missing file, or one that yields no usable proxies, prints a warning — except the implicit default (`proxies.txt` when `--proxies` is omitted), whose absence stays silent as the normal no-proxy state. |
 | `--version` | flag | — | Print the version and exit. Works even without `--urls`/`--output`. |
 
 `--urls` and `--output` are required. The parser accepts integer values for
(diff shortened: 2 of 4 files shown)


LeyckerS (approved):
Merging. This is the standard for a documentation PR here.

**You checked, and I checked your checking.** Every factual claim holds:

- `_PROXY_POOL.next()` appears at exactly one call site outside the tests — `moon_download.py:469`.
- `grep -n "proxy-server\|proxy=" moon_extract.py` returns **nothing**. The datanodes Chrome really is launched without one.
- `_CurlSession(impersonate=FF_IMPERSONATE, timeout=30, max_clients=64)` — no proxy argument, so the fuckingfast session is direct too.

The second and third are yours, not the issue's. #112 said "the browser goes direct"; you established that the `curl_cffi` path does as well, which makes it true of *both* providers rather than just the visible one. That is the kind of widening that makes a docs change worth more than the issue asked for.

**The framing is what I would have struggled to write.** This line, in particular:

> Rotation is bandwidth cover, not identity cover

That is the whole issue in six words, and it is the sentence that stops the next person filing this as a bug. Likewise putting the symptom in all four documents — *pages open, downloads fail, that is the design working* — because that is the exact observation that produced #112 in the first place.

Four files, consistent wording, no code touched, and you said so rather than leaving me to check. Closes #112.

One thing for whoever gets to it next, not for you: `docs/FAQ.md` still says *"The status bar shows how many were loaded"* a few lines above your addition. That is only true **after** a run has started — before then the chip reads "no proxy" regardless. It is #115, and your section makes the surrounding text slightly more misleading than it was, through no fault of yours.

Thanks, and welcome.

PR #129, opened 2026-08-02
ci: test supported Python versions through 3.14

### LINKED ISSUE #96: ci: the project promises Python 3.10+ but tests only 3.10-3.12 — including the version the maintainer actually runs

Vam-si-krish:
## Summary

Extends the lint workflow's Python matrix through 3.14, matching the supported Python range documented by the project. The existing `ruff.toml` target remains `py310`; this change only exercises the supported runtimes in CI.

Closes #96

## Validation

- Parsed `.github/workflows/lint.yml` and verified the matrix is exactly 3.10, 3.11, 3.12, 3.13, and 3.14.
- `ruff check .` (all checks passed)
- `pytest tests/ -q` (23 passed; one existing `aiohttp.BasicAuth` deprecation warning)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
     steps:
       - name: Checkout
         uses: actions/checkout@v7


LeyckerS (approved):
Three characters, and it answers the question #96 actually asked.

The issue said to add the versions and then see what breaks, because that was the unknown — the project promises 3.10+ and two of the versions covered by that promise had never been exercised. Your run settles it: **`syntax-check` is green on 3.10, 3.11, 3.12, 3.13 and 3.14**, with `fail-fast: false` so each reported independently rather than one masking the others.

Nothing broke, so there is no follow-up issue to open. The promise in the README is now backed by something.

You also left `ruff.toml`'s `target-version = "py310"` alone, which is right and which the issue asked for explicitly — it answers "does the syntax stay valid on the oldest supported version", a different question from "does it run on the newest". Changing it to track the matrix would have quietly dropped the floor.

Worth noting for the record: the maintainer develops on 3.14.6, so until this merged, every local run was on a version CI never touched. That gap is the one this closes.

Closes #96. Second one from you today.

PR #135, opened 2026-08-03
chore(deps): bound runtime dependencies and add tested constraints

### LINKED ISSUE #27: chore(deps): add upper bounds to requirements.txt and ship a constraints.txt

PomPomSaturin:
## Summary

* add major-version upper bounds to each runtime dependency in `requirements.txt`
* add `constraints.txt` with the current tested direct dependencies and the transitive packages requested in #27
* document constrained and unconstrained installation modes
* set Dependabot to `increase-if-necessary` so it does not keep raising already-compatible lower bounds
* add CI coverage for upper-bound enforcement, unconstrained installation, constrained installation, and `pip check`
* apply the same bounds to the no-Chrome test job

## Bounds

* `aiohttp>=3.9,<4`: preserves the documented minimum while excluding the next breaking major
* `playwright>=1.40,<2`: preserves the existing minimum and excludes a future major API break
* `curl_cffi>=0.7,<1`: keeps the current impersonation API family and excludes a breaking 1.x release

## Constraints

* `aiohttp==3.14.2`
* `playwright==1.61.0`
* `curl_cffi==0.15.0`
* `certifi==2026.7.22`
* `cffi==2.1.0`
* `greenlet==3.5.3`

## Validation

The new `dependency-contract` job performs both clean installation modes in separate virtual environments and runs `pip check` for each. It also fails if a runtime requirement is later added without an upper bound.

The existing no-Chrome suite remains unchanged apart from installing the bounded dependency ranges.

Closes #27

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,7 @@ updates:
       day: monday
       time: "09:00"
       timezone: Europe/Rome
+    versioning-strategy: increase-if-necessary
     open-pull-requests-limit: 3
     labels:
       - dependencies
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -5,10 +5,16 @@ on:
     branches: [main]
     paths:
       - "**.py"
+      - "requirements.txt"
+      - "constraints.txt"
+      - ".github/dependabot.yml"
       - ".github/workflows/lint.yml"
   pull_request:
     paths:
       - "**.py"
+      - "requirements.txt"
+      - "constraints.txt"
+      - ".github/dependabot.yml"
       - ".github/workflows/lint.yml"
 
 jobs:
@@ -36,6 +42,44 @@ jobs:
       - name: Lint with ruff
         run: ruff check .
 
+  dependency-contract:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Set up Python
+        uses: actions/setup-python@v7
+        with:
+          python-version: "3.10"
+
+      - name: Verify every runtime dependency has an upper bound
+        run: |
+          python - <<'PY'
+          from pathlib import Path
+
+          missing = []
+          for raw_line in Path("requirements.txt").read_text().splitlines():
+              line = raw_line.split("#", 1)[0].strip()
+              if line and "<" not in line:
+                  missing.append(line)
+
+          if missing:
+              raise SystemExit(f"runtime dependencies without upper bounds: {missing}")
+          PY
+
+      - name: Install latest compatible dependencies
+        run: |
+          python -m venv /tmp/moon-unconstrained
+          /tmp/moon-unconstrained/bin/pip install -r requirements.txt
+          /tmp/moon-unconstrained/bin/pip check
+
+      - name: Install the tested constrained set
+        run: |
+          python -m venv /tmp/moon-constrained
+          /tmp/moon-constrained/bin/pip install -r requirements.txt -c constraints.txt
+          /tmp/moon-constrained/bin/pip check
+
   no-chrome:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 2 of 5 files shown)


LeyckerS:
@PomPomSaturin — you closed this two minutes after opening it, with no comment. If that was deliberate, no problem at all and I will not press. But **if it was a misclick, please reopen it**, because this is complete work and I would merge it as it stands.

I read the whole diff before writing this. It does everything #27 asked and both things I added in the comments afterwards:

- **The bounds are back without the floors moving.** `aiohttp>=3.9,<4`, `playwright>=1.40,<2`, `curl_cffi>=0.7,<1`. Keeping the lower bounds where they were is the part that matters — I closed three Dependabot PRs last week precisely because raising a floor forbids old versions rather than permitting new ones, and it would have been easy to "tidy" them upward here.
- **`versioning-strategy: increase-if-necessary`** stops those PRs arriving again. That was in scope on the issue and most people would have skipped it.
- **The `dependency-contract` job is the half I care about most.** #27's history is that the bounds were added once in #12 and then quietly lost in `6a3dfac`, an unrelated engine rewrite that touched `requirements.txt` for another reason. Re-adding them fixes today; a CI job that **fails when any requirement has no upper bound** is what stops it happening a third time. You built exactly that.
- **`lint.yml`'s `paths:`** now includes `requirements.txt`, `constraints.txt` and `dependabot.yml`, so a change to any of them actually runs CI. Without it the new job would have been invisible on the pull requests that most need it.
- **The `no-chrome` job's inline `pip install` was updated too.** That line duplicates the requirements and is the kind of thing that silently drifts; catching it is the difference between a fix and a complete fix.

CI is green on all eight checks, including both `pip check` installs.

Reopen whenever you like — the button is on this page — or say the word and I will. Either way, `constraints.txt` and that contract job are worth having.

PomPomSaturin:
Hello!

That was my fault, I accidentally deleted the repo.  I will fix everything shortly, my fault again.

PR #136, opened 2026-08-04
chore(deps): bound runtime dependencies and add tested constraints

### LINKED ISSUE #27: chore(deps): add upper bounds to requirements.txt and ship a constraints.txt

PomPomSaturin:
This replaces #135, which GitHub cannot reopen because its original source repository was deleted.

Closes #27

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,7 @@ updates:
       day: monday
       time: "09:00"
       timezone: Europe/Rome
+    versioning-strategy: increase-if-necessary
     open-pull-requests-limit: 3
     labels:
       - dependencies
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -5,10 +5,16 @@ on:
     branches: [main]
     paths:
       - "**.py"
+      - "requirements.txt"
+      - "constraints.txt"
+      - ".github/dependabot.yml"
       - ".github/workflows/lint.yml"
   pull_request:
     paths:
       - "**.py"
+      - "requirements.txt"
+      - "constraints.txt"
+      - ".github/dependabot.yml"
       - ".github/workflows/lint.yml"
 
 jobs:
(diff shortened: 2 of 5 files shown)


LeyckerS (approved):
That explains the close — thank you for coming back and redoing it. Merging.

I read the diff in full on #135 and this is byte-identical, so nothing here is unreviewed. Restating why it goes in as it stands:

**The bounds are back and the floors did not move.** `aiohttp>=3.9,<4`, `playwright>=1.40,<2`, `curl_cffi>=0.7,<1`. That restraint is the point — I closed three Dependabot pull requests last week because raising a floor *forbids* old versions rather than permitting new ones, and tidying them upward here would have been easy and wrong.

**`versioning-strategy: increase-if-necessary`** stops those arriving again. It was in scope on the issue and it is the kind of line that gets skipped.

**The `dependency-contract` job is the half that matters most.** This has happened before: #12 added upper bounds and `6a3dfac` — an engine rewrite touching `requirements.txt` for an unrelated reason — removed them without anyone noticing. Re-adding them fixes today. A CI job that **fails when any requirement lacks an upper bound** is what stops a third occurrence. Both `pip check` installs, constrained and unconstrained, are a good addition on top.

**Two details that show you read the repository rather than just the issue:** extending `lint.yml`'s `paths:` to cover `requirements.txt`, `constraints.txt` and `dependabot.yml` — without it the new job would be invisible on exactly the pull requests that need it — and updating the inline `pip install` in the `no-chrome` job, which duplicates the requirements and is precisely where this sort of thing drifts.

Closes #27. Welcome, and sorry the fork deletion cost you a round trip.

Issue #84, opened 2026-08-01
the same speed is computed in MiB/s in some places and MB/s in others, all labelled "MB/s"


LeyckerS:
Speeds are divided by two different constants depending on where they are computed, and every one of them reaches the user labelled `MB/s`:

| Where | Code | Unit it actually is |
|:--|:--|:--|
| Per-file **live** speed | `moon_download.py:513` — `/ 1_048_576` | MiB/s |
| Per-file **final** average | `moon_download.py:545` — `/ 1e6` | MB/s |
| Engine **aggregate** speed | `moon_engine.py:582` — `/ 1_048_576` | MiB/s |
| Report **session speed** | `moon_download.py:275`, `337` — `/ 1e6` | MB/s |
| Stall detector threshold | `moon_download.py:525` — `/ 1e6` | MB/s |

`1_048_576 / 1_000_000 = 1.048576`, so the two families disagree by **4.9%** for the same real transfer rate.

Two visible consequences:

- A file downloading at a perfectly steady rate shows one number while it runs and a ~4.9% **higher** number the moment it completes, with nothing having changed.
- The live speed in the GUI and the "Session speed" line in the `moontech_*.log` report disagree by 4.9% for the same run.

The stall detector is the one place where the choice actually matters beyond display: `STALL_MIN_MBS` is compared against an `/1e6` value, so whichever convention is chosen, that comparison must stay consistent with the constant's intended meaning.

## Acceptance criteria

- One convention throughout. `/1e6` with the label `MB/s` is the honest pairing and matches what the reports already claim
- The live number and the final average for the same file agree
- `STALL_MIN_MBS` still means what its name and its comment say
- Say in the pull request which convention you picked and list every line you changed

No Windows machine needed to reason about it, though it is worth a real run to confirm the numbers now line up.

happykawayigt:
I can take this. I’ll standardize the live and aggregate paths on decimal MB/s, update the ETA conversion consistently, and add focused regression coverage.

LeyckerS:
@happykawayigt — yours, assigned.

Your plan of standardising on decimal MB/s is the right call: it is what the reports already claim in their own labels, so it is the convention that makes the fewest things liars.

Two things to watch:

- **`STALL_MIN_MBS` is not a display value.** It is compared against an `/1e6` figure in the stall detector, so whatever you do elsewhere, that comparison has to keep meaning what the constant's name says. Changing the units under it would quietly re-tune when a download gets killed.
[...]

PR #142, opened 2026-08-05
fix: live speed divides by the observation window instead of the sample span (#86)

### LINKED ISSUE #86: the per-file live speed in the GUI is roughly nine times the real rate
Measured on a real 82-link fuckingfast run, comparing what the GUI displayed against what the run's own report recorded.

| | Per-file speed |
|:--|:--|
| Shown live in the GUI, on every row | **~38 MB/s** |
| Recorded in `moontech_*.log` for the same run | **median 4.2 MB/s** (min 1.2, max 5.8, n=81) |

Roughly **nine times** overstated.

The report figure is the trustworthy one: `moon_download.py:545` computes it as `net / dl_s / 1e6` — total bytes of the file divided by its total download time. There is nothing to distort.

The arithmetic closes on the report side and not on the display side. Session speed was **172.0 MB/s** with a median download time of 120.6s across 81 files in a 242.5s run — about **40 files in flight on average**, and 40 × 4.2 ≈ 168 MB/s. The concurrency is working exactly as configured; only the number on screen is wrong.

## Cause

`moon_download.py:508-513`:

```python
pub_cut = now - 3.0
while pub_win and pub_win[0][0] < pub_cut:
    pub_win.popleft()
pub_span = max(now - pub_win[0][0], 0.25) if pub_win else 1.0
rec.live_mbs = sum(b for _, b in pub_win) / pub_span / 1_048_576
```

`pub_span` is the age of the **oldest surviving sample**, not the length of the observation window. Those are the same number only while chunks arrive continuously.

TCP delivery is bursty. After a quiet stretch, the old samples get popped and what remains may span a fraction of a second while still holding a full burst of bytes. Dividing a burst by 0.25s yields the burst's peak rate, and that is what gets published. The floor of `0.25` puts a hard ceiling of 4× on how badly a single sample can distort it, but repeated bursts keep the published value pinned near peak instead of near average.

The aggregate speed in `moon_engine.py:582` has the same shape with a `0.05` floor, so it is distorted the same way — the operator reported 220-280 MB/s on screen against a recorded session speed of 172.0 MB/s.

## The fix

Divide by the observation window, not by the sample spread. The denominator should be the smaller of the window length (3.0s) and the time since the download started — never the gap between retained samples.

## Acceptance criteria

- On a steady download, the live per-file number and the file's final average in `moontech_*.log` agree within a few percent
- The live aggregate agrees with the report's `Session speed` line to the same tolerance
- The readout still reacts within about a second to a genuine change in rate — the point is to stop it reading peak instead of average, not to make it sluggish
- Verify on a real run and put the before/after numbers in the pull request

Worth doing together with #84, which is the unit mismatch in the same expressions.

Found by running the program.

XEDAB:
Fix #86

The live speed used the age of the oldest retained sample as its denominator, which measures the span of the bursts currently buffered rather than the observation window itself. On bursty TCP delivery a 3s window can collapse to a fraction of a second, and the floor (0.25s / 0.05s) publishes the burst's peak as the rate — the ~9x overstatement reported in #86.

Divide by the window length instead: min(3.0s, time since start), so steady downloads read the average and sub-3s runs use the elapsed time. Applied to the per-file live speed (moon_download.py), the engine aggregate (moon_engine.py) and the CLI progress copy (moon_cli.py). Unit conversion (/1_048_576) left untouched — that mismatch belongs to #84.

Adds tests/test_snapshot_speed.py covering the bursty-window and sub-3s fallback.

## Type of change

- [ ] Documentation update
- [ ] Refactor / code cleanup
- [ ] Other:

## Checklist

      description

图片
图片
图片
图片


XEDAB:
One less blank line and the CI blows up...
8fc2dd738bf174a0b360f6912badcae76478845

LeyckerS (approved):
This is the fix. Merging — and you do not need luck, the test you wrote does the job you were worried about.

**You cannot run it against a real download, but the arithmetic is checkable and it checks out.** With the three samples in your test and `t0` a thousand seconds back, the window is the full 3.0s: 3 MB ÷ 3.0 ÷ 1048576 ≈ 0.95 MB/s, inside your `0.5 < x < 1.5`. With `t0` 1.5s back the window shortens to 1.5s and it reads ≈1.91, inside `1.5 < x < 2.5`. **On the old code the denominator would have been 0.20s** — the age of the oldest retained sample — giving ≈14.3 MB/s, outside both bounds. So it fails before and passes after, which is what a regression test has to do.

Covering both the full-window and the short-run case in one test is the part I would have asked for.

**All three call sites, including the one that keeps getting forgotten.** `moon_engine.py` for the aggregate against `_t0`, `moon_download.py` for the per-file figure against `dl_t0`, and `moon_cli.py:111`, which carries its own copy of this expression because the CLI keeps separate progress accounting (#97). A fix that touched only the engine would have left the CLI reporting different numbers for the same run.

**One behaviour change worth knowing**, not a defect: for the first three seconds after downloads begin, the displayed rate now ramps rather than jumping. If bytes have only been arriving for half a second, dividing by 3.0 gives the true average over the last three seconds — 2.5 of which were idle. That is the correct reading of a windowed average and it is what stops the peak being published, but the first few refreshes of a run will look lower than they used to.

**Two small things, neither worth another round:**

`max(min(3.0,now - t0),0.05)` is missing spaces after the commas, in all three places. Nothing catches it — ruff's whitespace rules are preview-only and `ruff.toml` does not enable them — but the surrounding code has them.

You left the `1_048_576` alone. That is right: the MiB/MB mismatch is #84 and it is assigned to @happykawayigt. Worth a line in the description next time saying you left it deliberately, so a reviewer does not wonder.

The number that produced this issue came from a real run — 38 MB/s on screen against a median 4.2 in the same run's report. The maintainer can reproduce that setup, so we will have a real measurement to confirm against shortly. I will report back here either way.

Closes #86. Third merged from you.

PR #143, opened 2026-08-05
engine/cli: collect and report browser worker failures

### LINKED ISSUE #59: engine/cli: worker gather is fail-fast while the straggler gather is not, and nothing says why
### Context

Two `asyncio.gather` calls launch the browser workers, and neither passes `return_exceptions=True`:

```python
# moon_engine.py:348
await asyncio.gather(*[asyncio.create_task(_launch(i)) for i in range(n_workers)])

# moon_cli.py:235
await asyncio.gather(*[asyncio.create_task(browser_worker(gate.get, i))
                       for i in range(n_workers)])
```

Ten lines below each of them, the straggler gather does pass it:

```python
# moon_engine.py:358
await asyncio.gather(*stragglers, return_exceptions=True)
```

So the same file treats the same hazard two different ways, twenty lines apart, and nothing explains why.

Without `return_exceptions=True`, `gather` re-raises the first exception as soon as it appears. The sibling worker tasks are neither cancelled nor awaited — they keep running while the `finally` block tears the browser gate down underneath them (`moon_engine.py:349-352`), and the run ends without its summary. With it, every worker is awaited and the failures come back as results you can log per worker.

**Worth stating honestly:** I could not construct a case where this fires today. `_browser_worker` handles its own failures and `download_file` catches broadly (`moon_download.py:562`), so in practice a worker returning normally is the only observed outcome. This is about the pattern being inconsistent and load-bearing, not about a reproducible crash.

### What to do

Pick one of two outcomes — either is acceptable, but the PR has to say which and why:

**Either** make the worker gathers consistent with the straggler ones: pass `return_exceptions=True`, then inspect the returned list and log any exception against its worker id so a failure is visible instead of silent.

**Or** leave the behaviour as-is and add a comment at both sites explaining that a worker is not expected to raise and that a fail-fast teardown is the intended response if it does. `CONTRIBUTING.md` asks for exactly this when a swallow or a sharp edge is deliberate.

What is not acceptable is leaving it undocumented and inconsistent.

If you take the first option, do not add retry logic, do not change the teardown order, and do not touch the straggler gathers.

### Acceptance criteria

- [ ] `moon_engine.py:348` and `moon_cli.py:235` are handled the same way as each other
- [ ] The PR states which of the two outcomes was chosen and why
- [ ] If exceptions are collected, each one is logged with the worker id that produced it
- [ ] A worker raising does not leave the browser gate closed underneath running siblings
- [ ] `pytest tests/ -q` passes
- [ ] No change to the two straggler gathers

### Notes for the contributor

**No Windows needed**, but this is **not a first issue.** You need to be comfortable reading `asyncio` teardown and reasoning about task lifetimes, and the valuable part is the argument in the PR description, not the diff.

A test is welcome but optional: stubbing a worker that raises, then asserting the other workers were awaited, is genuinely fiddly and I would rather have a clear explanation than a brittle test.

PomPomSaturin:
## Summary

- make the engine and CLI browser-worker gathers use `return_exceptions=True`
- wait for every sibling worker before closing `BrowserGate`
- report each collected exception with the worker id that produced it
- leave retries, teardown ordering, and both straggler gathers unchanged

## Why this approach

I chose collection rather than documenting the existing fail-fast behavior. A worker is not expected to raise today, but if one does, the old gather raised immediately and allowed `finally` to close the shared browser gate while sibling workers could still be running.

Collecting results keeps both front-ends consistent and ensures the gate is only closed after all worker tasks have finished.

The result list preserves task order, so enumerating it maps each exception back to the same worker id used when the task was created.

## Validation

- `python -m compileall -q moon_engine.py moon_cli.py`
- `pytest tests/ -q`
- verified the branch is one commit and changes only `moon_engine.py` and `moon_cli.py`

Closes #59

### DIFF
--- a/moon_cli.py
+++ b/moon_cli.py
@@ -249,8 +249,13 @@ async def run(urls: list[str], output_dir: str, n_workers: int,
         on_open=lambda: print("  [chrome] datanodes link found - starting Chrome", flush=True))
 
     try:
-        await asyncio.gather(*[asyncio.create_task(browser_worker(gate.get, i))
-                               for i in range(n_workers)])
+        worker_results = await asyncio.gather(
+            *[asyncio.create_task(browser_worker(gate.get, i)) for i in range(n_workers)],
+            return_exceptions=True,
+        )
+        for wid, result in enumerate(worker_results):
+            if isinstance(result, BaseException):
+                print(f"  [worker {wid} error] {result}")
     finally:
         await gate.aclose()
 
--- a/moon_engine.py
+++ b/moon_engine.py
@@ -376,7 +376,13 @@ class Engine:
                 output_links, failed_urls, dest_folder, mode, max_retries, telem)
 
         try:
-            await asyncio.gather(*[asyncio.create_task(_launch(i)) for i in range(n_workers)])
+            worker_results = await asyncio.gather(
+                *[asyncio.create_task(_launch(i)) for i in range(n_workers)],
+                return_exceptions=True,
+            )
+            for wid, result in enumerate(worker_results):
+                if isinstance(result, BaseException):
+                    self.log(f"  ✗  worker {wid} failed: {result}", "fail")
         finally:
             if browser_started:
                 self._inc("_browsers", -1)


LeyckerS (approved):
Merging. You picked the harder of the two options the issue offered and the reasoning in your description is the right one.

#59 allowed either answer — make the worker gather match the straggler gather, or document why they differ. Collecting is better, and for the reason you gave: a worker raising is not expected today, but if one ever does, the old `gather` propagated immediately and the siblings were left to be torn down mid-flight with nothing said about which one failed or why. `return_exceptions=True` plus the worker id turns a silent abort into a line naming the culprit.

Both front-ends, which is the standard here and the thing most often missed. And you left both straggler gathers alone — they already collected, so touching them would have been churn.

The `isinstance(result, BaseException)` check rather than `Exception` is the correct choice: `gather` returns `CancelledError` in that list too, and it does not inherit from `Exception` in modern Python. Catching only `Exception` would have let a cancelled worker pass silently.

Closes #59. Second merged from you, and the `dependency-contract` job you added yesterday has already validated four dependency pull requests since.

PR #146, opened 2026-08-05
cli: add --extractors with deprecated --browsers alias

### LINKED ISSUE #45: cli: --browsers does not control browsers — rename to --extractors with a deprecated alias

PomPomSaturin:
## Summary

- add `--extractors` as the primary CLI flag for parallel extraction workers
- keep `--browsers` working as a deprecated compatibility alias
- hide the deprecated alias from `--help`
- prefer `--extractors` and print a warning when both flags are supplied
- update the README, quickstart, and CLI reference examples
- add focused tests for the new flag, deprecated alias, precedence, and help output

## Behavior

`--extractors N` is now the advertised option and keeps the existing default of 8.

`--browsers N` still works for existing scripts but prints:

```text
WARNING: --browsers is deprecated; use --extractors instead
```

When both are passed, --extractors wins and the CLI reports that the deprecated value was ignored.

This does not change browser lifecycle behavior. BrowserGate still decides when the one shared Chrome instance is needed; the flag controls extraction workers only.

Validation
installed the constrained runtime dependency set
python -m compileall -q moon_cli.py tests/test_cli_extractors_flag.py
pytest tests/ -q
git diff --check
verified the branch contains one commit and only the five intended files

Closes #45

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,7 +270,7 @@ Full write-up: **[docs/ARCHITECTURE.md](docs/ARCHITECTURE.md)**.
 | [Architecture](docs/ARCHITECTURE.md) | how the engine is built, feature by feature |
 | [Engineering notes](docs/ENGINEERING_NOTES.md) | the measurements behind the design decisions |
 | [Troubleshooting](docs/TROUBLESHOOTING.md) | 403s, Turnstile failures, CDP conflicts, stalls |
-| [FAQ](docs/FAQ.md) | why curl_cffi is mandatory, what `--browsers` means |
+| [FAQ](docs/FAQ.md) | why curl_cffi is mandatory, what `--extractors` means |
 | [Contributing](CONTRIBUTING.md) | architecture, what counts as a contribution, the verification commands |
 | [Changelog](CHANGELOG.md) | every version since 14.0 |
 
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -17,14 +17,16 @@ are ignored.
 |:--|:--|:--|:--|
 | `--urls` | path | required | Text file containing one URL per line. The CLI exits before a run if the file is missing or has no usable URLs. |
 | `--output` | path | required | Directory where downloaded files are written. It is created if needed. |
-| `--browsers` | integer | `8` | Number of parallel extraction workers. Despite the name, it does not start that many browsers: datanodes uses one shared Chrome instance. |
+| `--extractors` | integer | `8` | Number of parallel extraction workers. Despite the name, it does not start that many browsers: datanodes uses one shared Chrome instance. |
+
+`--browsers` remains accepted as a deprecated compatibility alias. It is hidden from `--help`; when both flags are supplied, `--extractors` takes precedence and the CLI prints a warning.
 | `--streams` | integer | `24` | Maximum number of concurrent download streams. |
 | `--retries` | integer | `3` | Maximum extraction attempts per URL. Network retries inside one download are separate. |
 | `--proxies` | path | `proxies.txt` | Proxy-list file to load. Applies to **downloads only** — extraction always goes direct, see below. A missing file, or one that yields no usable proxies, prints a warning — except the implicit default (`proxies.txt` when `--proxies` is omitted), whose absence stays silent as the normal no-proxy state. |
 | `--version` | flag | — | Print the version and exit. Works even without `--urls`/`--output`. |
 
 `--urls` and `--output` are required. The parser accepts integer values for
-`--browsers`, `--streams`, and `--retries`; it does not add further CLI-side range
+`--extractors`, `--streams`, and `--retries`; it does not add further CLI-side range
 validation.
 
 The parser's actual default for `--proxies` is unset (`None`); `proxies.txt` is the
@@ -61,7 +63,7 @@ https://fuckingfast.co/example-b#example-b.zip
 python moon_cli.py \
   --urls fuckingfast-links.txt \
   --output ./downloads/fuckingfast \
-  --browsers 8 \
+  --extractors 8 \
   --streams 24
 ```
 
@@ -80,7 +82,7 @@ https://datanodes.to/example-b
 python moon_cli.py \
   --urls datanodes-links.txt \
   --output ./downloads/datanodes \
-  --browsers 8 \
+  --extractors 8 \
   --streams 24
 ```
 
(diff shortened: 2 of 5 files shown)


LeyckerS (changes requested):
The implementation is right and you handled the trap the roadmap warned about. One rendering problem in the docs, then this merges.

## Blocking: the flags table breaks in half

In `docs/CLI.md` the alias paragraph lands **inside** the table:

```
| `--extractors` | integer | `8` | Number of parallel extraction workers... |
                                                          <- blank line
`--browsers` remains accepted as a deprecated compatibility alias...
| `--streams` | integer | `24` | Maximum number of concurrent download streams. |
```

A Markdown table ends at the first blank line. Everything after that paragraph — `--streams`, `--retries`, `--proxies`, `--version` — has no header and no delimiter row, so it renders as literal text full of pipe characters rather than as four table rows. Four of the seven documented flags become unreadable.

Move the paragraph below the table, before the `--urls and --output are required` line. Nothing else about it needs to change.

## While you are in there

The `--extractors` row still carries the old explanation:

> Number of parallel extraction workers. **Despite the name, it does not start that many browsers:** datanodes uses one shared Chrome instance.

That clause existed because the name was wrong. It is now a leftover that reintroduces the confusion the rename removes — a reader meets `--extractors`, is told not to assume browsers, and wonders why anyone would have. The useful half is the fact: *datanodes uses one shared Chrome instance regardless of this value.* Keep that, drop the apology for the name.

## What is right, and it is most of it

**You solved the `argparse.SUPPRESS` problem correctly.** #45 flagged that `tests/test_docs_cli_flags.py` reads the real flag set from `--help`, so hiding the alias would fail any documentation still showing `--browsers` in a command. You hid it *and* moved every `python moon_cli.py …` example over, and `docs-cli-flags` is green. That was the trap and you walked round it.

**`default=None` on both, resolved afterwards.** The only way to tell "user passed 8" from "user passed nothing" is a sentinel, and without it the precedence rule could not exist. Straightforward once seen and easy to get wrong.

**Both-flags precedence with a distinct message** rather than silently picking one.

**The tests cover the four cases that matter** — primary flag silent, alias warns, `--extractors` wins, alias absent from `--help` — and putting the resolution block before the `--urls` existence check is what makes them observable without a real file. Deliberate, and it works.

Third from you, and the first two are already earning: the `dependency-contract` job from #136 has validated four dependency pull requests, and #143 is what will name the worker the next time one fails.

PomPomSaturin:
Addressed in 9c2c43e.

- moved the deprecated `--browsers` paragraph below the full arguments table so the remaining rows render correctly
- simplified the `--extractors` description and kept only the useful note that datanodes uses one shared Chrome instance regardless of this value

No other files were changed.

LeyckerS (approved):
Both fixed exactly. Merging.

The paragraph now sits below the full table, so all seven flag rows render. And the description became the useful half:

> Number of parallel extraction workers. Datanodes uses one shared Chrome instance regardless of this value.

That is the fact a reader needs, without the apology for a name that is no longer wrong.

This closes #45, which has been open since before any of the current contributors arrived. The `--browsers` name has cost every new user the same ten minutes of working out that it does not start browsers, and the deprecated alias means no existing script breaks on the way out.

Worth recording what you navigated, because it was the trap on this issue: `tests/test_docs_cli_flags.py` reads the real flag set from `moon_cli.py --help`, so hiding the alias with `argparse.SUPPRESS` makes any documentation still showing `--browsers` in a command fail the build. You hid it *and* moved every example, and `docs-cli-flags` is green — which is the check proving the docs and the parser agree.

Third merged from you in two days, and the first two are still working: `dependency-contract` has validated five dependency pull requests, and #143 is what will name the worker the next time one fails.

PR #150, opened 2026-08-06
fix: stop downloads when destination is full (#116)

### LINKED ISSUE #116: a full disk is treated as an ordinary per-file error: the run continues and burns bandwidth it cannot write
When the destination fills up, every remaining file is still extracted, downloaded in full, and only then fails on the write. The run does not stop and the user is not told.

## Observed

58 links, destination drive with no free space left:

```
  Total links    : 58
  Completed OK   : 12
  Failed         : 46
  Total data     : 4.36 GB
  Median DL time : 52.9s
```

The GUI meanwhile reported **16.58 GB** transferred over 4m 50s. Every one of the 46 failures carries the same note:

```
  1    Halo_Campaign_Evolved_...    0   0    0.0    0.4   236.1    —  fail
       → error att 1: [Errno 28] No space left on device
  2    Halo_Campaign_Evolved_...    1   0    0.0    0.5   236.5    —  fail
       → error att 1: [Errno 28] No space left on device
  3    Halo_Campaign_Evolved_...    2   0    0.0    0.3   236.2    —  fail
```

Each of those spent **236 seconds downloading** before failing on a write that could never succeed. Roughly 12 GB pulled off the network and discarded.

## Cause

`ENOSPC` is not distinguished anywhere — `grep -n "ENOSPC\|No space" moon_download.py` returns nothing. It lands in the catch-all:

```python
except Exception as e:
    err = str(e)
    rec.notes.append(f"error att {att+1}: {err}")
    if att < DL_INNER_RETRIES - 1 and ("ContentLengthError" in err or "not enough data" in err.lower()):
        await asyncio.sleep(0.5 * (att + 1))
        continue
    return False, err, downloaded
```

It correctly avoids the inner retry, then returns a per-file failure and the worker moves to the next link — which will fail identically, because the disk does not empty itself between files.

The note goes to `rec.notes`, which only reaches the report after the run (#99), so nothing on screen ever says "disk full". The user sees files failing for no stated reason.

## What is needed

`ENOSPC` is not a per-file problem. It is a **run-level, non-recoverable** condition and the only correct response is to stop:

- Detect it specifically — `isinstance(e, OSError) and e.errno == errno.ENOSPC` rather than matching on the message, which is localised (the report above is from an Italian Windows and reads *"No space left on device"* only because Python normalises it; other paths do not)
- Abort the run rather than continuing to the next link
- Say so unmistakably, in the live log and not only in the report: which folder, how much was needed
- Leave the `.tmp` files alone — they are resumable once space is freed, and the run had 46 of them

Worth deciding and stating in the pull request: whether in-flight transfers are cancelled immediately or allowed to finish. Cancelling recovers the bandwidth sooner; finishing means fewer half-files. Either is defensible.

## Acceptance criteria

- A run that hits `ENOSPC` stops instead of attempting every remaining link
- The reason appears on screen while it happens, naming the destination folder
- Existing `.tmp` files are preserved for resume
- Detection is by `errno`, not by string matching
- A regression test with a write that raises `OSError(errno.ENOSPC, ...)` — no network needed, `tests/test_resume_200.py` shows the pattern for faking the response surface

## Environment

Windows 10, Python 3.14.6, destination on a drive with 0 bytes free of 931 GB. Found by running the program.

aislopking024:
## Description

Fixes #116.

A destination filesystem running out of space is now treated as a run-level fatal condition instead of an ordinary per-file failure. The first `ENOSPC` event publishes a shared fatal signal, stops new queue intake and retries, and lets active transfers unwind cooperatively at their next stream/write boundary.

The triggering and interrupted transfers keep their `.tmp` files. Only the triggering URL is added to the failed list; peers interrupted by the shared fatal state are classified as aborted. Both the engine and CLI await their active work, attempt telemetry/report persistence, contain secondary persistence failures, and finish with disk-full-specific messaging.

Cooperative unwinding was chosen over immediate task cancellation because active file writes run in executor threads and cannot be safely cancelled mid-write.

## Root cause

The downloader's catch-all per-file error path classified `OSError(errno.ENOSPC)` like a normal transfer error, allowing the queue and retry machinery to continue consuming bandwidth for data that could not be persisted.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- `pytest tests/test_disk_full.py -q` — 13 passed (run twice before rebase; passed again after integration)
- `pytest tests/ -q` — 42 passed
- `ruff check moon_download.py moon_engine.py moon_cli.py tests/test_disk_full.py tests/test_exit_cleanup.py` — passed
- `git diff --check` — passed

The network-free regressions cover partial and short writes, exact remaining-byte reporting, active-peer exception races, retry/backoff suppression, retry-list and telemetry failures, terminal engine state, both frontends, and integration with #149's in-flight stop handling.

## Disclosure

Implementation and review were assisted by OpenAI Codex using GPT-5.6 Terra and GPT-5.6 Sol.

### DIFF
--- a/moon_cli.py
+++ b/moon_cli.py
@@ -19,6 +19,7 @@ from moon_download import (
     _close_sess,
     _sanitize_filename,
     download_file,
+    RunFatalControl,
 )
 
 # ── EXTRACTION ────────────────────────────────────────────────────────────────
@@ -76,6 +77,7 @@ async def run(urls: list[str], output_dir: str, n_workers: int,
     ok_count      = 0
     fail_count    = 0
     dls_active    = 0
+    fatal_control = RunFatalControl()
 
     cfg = {"browsers": n_workers, "dl_streams": max_dl, "retries": max_retries,
            "total_links": len(urls)}
@@ -101,7 +103,12 @@ async def run(urls: list[str], output_dir: str, n_workers: int,
     stop_progress = asyncio.Event()
     async def progress_loop():
         while not stop_progress.is_set():
-            await asyncio.sleep(2.0)
+            try:
+                await asyncio.wait_for(stop_progress.wait(), timeout=2.0)
+            except asyncio.TimeoutError:
+                pass
+            if stop_progress.is_set():
+                break
             snap = list(bytes_acc)
             now  = time.monotonic()
             cut  = now - 3.0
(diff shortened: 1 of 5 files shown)


LeyckerS (approved):
Read the diff against #116's acceptance criteria rather than the description, and all five hold:

- **errno-based detection** — `moon_download.py`, `except OSError as e: if e.errno == errno.ENOSPC:`. Not a string match on the message, which is what the issue asked for.
- **The run aborts** — `RunFatalControl` holds the first `DiskFullError` under a lock and both `_do_dl` and `_browser_worker` check `fatal_control.is_set()` before taking new work. Keeping this separate from the stall-kill state was the right call; overloading `kill_evt` would have made a full disk look like a stall and re-extract.
- **The log names the folder** — `Disk full in {folder}: need {n:,} bytes to continue`, plus the run-level summary line.
- **`.tmp` preserved** — confirmed by `os.path.getsize(tmp)` being read back to compute the shortfall; nothing unlinks it on this path.
- **Regression test** — thirteen of them.

Three things I want to call out because they are better than what the issue asked for:

`_do_dl` looks like a 60-line deletion in the diff, but it is the existing body re-indented into `try:` so that `self._inc("_dls",-1)` can move to a `finally:`. It was previously repeated on three exit paths — and this PR adds a fourth, which is exactly the situation where that pattern leaks the counter. Good catch.

The `memoryview` write loop is not padding either. `write()` can return a short count on a nearly-full disk instead of raising, and without the loop that silently truncates the file. `test_short_write_then_enospc_reports_actual_remaining_bytes` is the test that earns it.

Mapping `"aborted"` to `"queue"` in `_FILE_UI_STATE` rather than to `"fail"` is the detail I would have argued for. #116's actual complaint was that a full disk got reported as an ordinary per-file failure; marking each file `fail` would have reproduced the bug in the UI while fixing it in the engine.

One sequencing note, not a change request: #149 rewrites the same `moon_engine.py` region (lines ~175–200) for #65. Merging this first means that PR needs a rebase — that is on me to coordinate, not on you.

Merging.

LeyckerS:
@shard872 — this is approved and ready to go on my side: CI is green on all eight checks, `mergeable` is CLEAN, and I have checked the diff against #116's criteria one by one (details in the review above).

The only thing stopping the merge is that the PR is still marked **draft**, which I cannot flip for you. The body reads as finished work, so I assume it was an oversight — click "Ready for review" and I will squash it straight away.

If it is *not* an oversight and something is still outstanding, say so and I will hold. No rush either way.

LeyckerS:
Correcting myself: in the review above I said that merging this first would put the rebase on #149. It went the other way. This PR is still in draft and I cannot flip that flag, #149 was complete and CI-green, so I merged #149 and **this one now conflicts with `main`** (`CONFLICTING / DIRTY`). The rebase landed on you because of my sequencing, not because of anything in your work — sorry about that.

Here is exactly what moved under you, so you are not reverse-engineering it from a conflict marker.

**All of it is in `_do_dl` and `stop()`.** #149 fixed #65, where Stop waited for in-flight transfers to finish:

- `Engine.__init__` gained `self._active_kill_events = set()`, cleared again in `start()`.
- In `_do_dl`, `kill_evt` is registered in that set under `self._lock`, and the `await download_file(...)` call is wrapped in `try:` / `finally: … discard(kill_evt)`. **This is an inner `try:` around the single await, not around the whole body** — your outer `try:` / `finally: self._inc("_dls",-1)` goes around it and the two nest cleanly.
- A new branch was inserted into the chain:

```python
elif msg == "stall_killed" and self._get("_stop_flag"):
    rec.status = "stopped"
```

  It sits **above** the generic `elif msg == "stall_killed":`. Please keep that order when you resolve — reversed, a user stop is treated as a stall kill and re-queued, which is the bug #149 exists to fix. Your `elif msg == "aborted_disk_full":` can go directly after it; the two conditions are disjoint.

- `stop()` now signals those events with `loop.call_soon_threadsafe(kill_evt.set)`.

One thing that should make your diff *smaller*: I have already added `"stopped": "queue"` to `_FILE_UI_STATE` (`moon_engine.py:471`), so your `"aborted": "queue"` line now slots in beside an existing precedent instead of being the only non-obvious entry.

Nothing about the approval changes — the review stands, it is still a yes from me. When you have rebased, mark it ready and I will squash it.

LeyckerS:
Rebase checked — the branch order in `_do_dl` survived it intact:

```python
if ok: ...
elif msg == "stall_killed" and self._get("_stop_flag"):   # → "stopped"
elif msg == "aborted_disk_full":                          # → "aborted"
elif msg == "stall_killed":                               # → re-queue
```

`aborted_disk_full` sitting between the two `stall_killed` arms is the right place for it: the conditions are disjoint, and the user-stop arm stays ahead of the generic one, so neither a stop nor a full disk can be re-queued by the retry path.

All eight checks green. Merging — thanks for the rebase, and for the patch.

PR #152, opened 2026-08-06
fix: use Proxy-Authorization header for proxy auth

### LINKED ISSUE #151: silence the aiohttp BasicAuth deprecation warning in the test run (the call itself has to stay)
> **This issue was rewritten on 7 August 2026.** The original version asked for the wrong fix and @slegarraga implemented it faithfully in #152 before I caught it. The reasoning and the evidence are in [this comment](https://github.com/LeyckerS/moondownloader/issues/151#issuecomment-5211514801); the short version is that moving proxy credentials into a header breaks them for https downloads, which is all of them. The scope below is what the issue should have said in the first place.

Every test run prints this:

```
tests/test_proxy_pool.py::test_proxy_pool_load_valid
  moon_download.py:114: DeprecationWarning: BasicAuth is deprecated and will be removed
  in aiohttp 4.0; use aiohttp.encode_basic_auth() with headers={'Authorization': ...} instead
    "auth": aiohttp.BasicAuth(user, passwd),
```

**The `BasicAuth` call is staying.** This issue is only about the noise in the test output.

## Why the call stays

`moon_download.py:114` builds proxy credentials, and they are handed to aiohttp as `proxy_auth`. That parameter accepts nothing else:

```python
if proxy_auth and not isinstance(proxy_auth, helpers.BasicAuth):
    raise ValueError("proxy_auth must be None or BasicAuth() tuple")
```

and `helpers.BasicAuth(...)` is what emits the warning. aiohttp has deprecated the only type its own `proxy_auth` parameter accepts, without shipping a replacement for that parameter.

Passing the credentials as a header instead is not a workaround — it changes what goes on the wire. For an https target aiohttp authenticates the proxy on a `CONNECT` request, so anything in `headers=` is sent inside the finished TLS tunnel where the proxy cannot read it. Both providers here serve https, so that breaks authenticated proxies on every real download while still passing every unit test. Do not do it.

`requirements.txt` pins `aiohttp>=3.9,<4`, so the upper bound holds the actual removal off. When aiohttp 4 ships a proxy-auth API, that is a separate issue and a real one.

## The work

Give pytest a `filterwarnings` entry so this specific warning does not print. The repo has **no pytest configuration file at all** today, so this means creating one — either `pytest.ini` or a `[tool.pytest.ini_options]` block in a new `pyproject.toml`. Pick whichever you can justify in the PR body; I have no preference beyond it being a deliberate choice.

Two things the entry must get right:

1. **Target this warning only.** A blanket `ignore::DeprecationWarning` would hide every future deprecation in every dependency, which is a worse outcome than the noise. Match on the message or the module.
2. **Leave a comment pointing at this issue.** Otherwise someone reads a suppressed warning in a year, "fixes" it properly, and re-introduces the bug described above. The comment is the actual deliverable here; the config line is two minutes.

## Done when

- `pytest tests/ -q` runs clean with no `BasicAuth` deprecation output
- An unrelated `DeprecationWarning` would still be shown — check by adding one temporarily, or by matching narrowly enough that it is obvious from reading
- `moon_download.py` is unchanged
- The config file explains why the suppression exists and links here

Small — genuinely small this time. Comment to claim it.

slegarraga:
Closes #151.

Builds proxy credentials as a `Proxy-Authorization` header value instead of `aiohttp.BasicAuth`. Uses `aiohttp.encode_basic_auth` when available and falls back to `BasicAuth(...).encode()` otherwise, so the `aiohttp>=3.9` floor does not change.

Verified:
- `pytest tests/ -q` -> 32 passed
- `python -W error::DeprecationWarning -m pytest tests/test_proxy_pool.py -q` -> 8 passed

### DIFF
--- a/moon_download.py
+++ b/moon_download.py
@@ -111,7 +111,7 @@ def parse_proxy_line(line: str) -> dict | None:
                 user, passwd, ip, port = parts
             return {
                 "url": f"http://{ip}:{port}",
-                "auth": aiohttp.BasicAuth(user, passwd),
+                "auth": _proxy_auth_header(user, passwd),
             }
         elif len(parts) == 2:
             ip, port = parts
@@ -119,6 +119,13 @@ def parse_proxy_line(line: str) -> dict | None:
     except Exception:
         pass
     return None
+
+
+def _proxy_auth_header(user: str, passwd: str) -> str:
+    """Build a Proxy-Authorization header value without deprecated BasicAuth."""
+    if hasattr(aiohttp, "encode_basic_auth"):
+        return aiohttp.encode_basic_auth(user, passwd)
+    return aiohttp.BasicAuth(user, passwd).encode()
 
 def count_usable_proxies(path: str) -> tuple[int, int]:
     if not os.path.exists(path):
@@ -495,7 +502,8 @@ async def download_file(
             req_kwargs = dict(headers=hdrs)
             if dl_proxy:
                 req_kwargs["proxy"] = dl_proxy
-                req_kwargs["proxy_auth"] = dl_proxy_auth
+                if dl_proxy_auth:
+                    hdrs["Proxy-Authorization"] = dl_proxy_auth
 
             async with dl_session.get(proxy_url, **req_kwargs) as r:
                 if r.status == 416:
--- a/tests/test_proxy_pool.py
+++ b/tests/test_proxy_pool.py
@@ -1,5 +1,26 @@
 import pytest
-from moon_download import ProxyPool
+from moon_download import ProxyPool, parse_proxy_line
+
+
+def test_parse_proxy_line_builds_proxy_auth_header():
+    cfg = parse_proxy_line("1.2.3.4:8080:user:pass")
+    assert cfg is not None
+    assert cfg["url"] == "http://1.2.3.4:8080"
+    assert cfg["auth"] == "Basic dXNlcjpwYXNz"
+
+
+def test_parse_proxy_line_builds_auth_header_for_credentials_first():
+    cfg = parse_proxy_line("user:pass:5.6.7.8:9090")
+    assert cfg is not None
+    assert cfg["url"] == "http://5.6.7.8:9090"
+    assert cfg["auth"] == "Basic dXNlcjpwYXNz"
+
+
+def test_parse_proxy_line_no_auth_for_two_part_line():
+    cfg = parse_proxy_line("5.6.7.8:9090")
+    assert cfg is not None
+    assert cfg["url"] == "http://5.6.7.8:9090"
+    assert cfg["auth"] is None
 
 
 def test_proxy_pool_load_valid(tmp_path):


LeyckerS (commented):
@slegarraga — first: you did what the issue asked, you did it quickly, you avoided the `Authorization`/`Proxy-Authorization` trap the issue warned about, you kept the `>=3.9` floor with a clean feature-detect instead of moving it, and you verified with `-W error::DeprecationWarning`. That is a better-executed first pull request than most. The problem is upstream of all of it: **the issue told you to do the wrong thing.** That is my mistake, and I have corrected it publicly on #151.

Here is what happens with this change on the path the program actually uses.

`aiohttp/connector.py`, `_update_proxy_auth_header_and_build_proxy_req`:

```python
if not req.is_ssl():
    req.headers[hdrs.PROXY_AUTHORIZATION] = auth        # http  → on the request
else:
    proxy_req.headers[hdrs.PROXY_AUTHORIZATION] = auth  # https → on the CONNECT
```

For an https target, aiohttp opens a `CONNECT` tunnel and the proxy authenticates *that*. `moon_download.py:502` now puts the credentials in `hdrs`, which becomes `headers=` on the tunnelled request — sent **inside** TLS, encrypted end to end with the file host. The proxy cannot read it and answers `407`.

Both providers hand out https links, so this affects every real download through an authenticated proxy while looking correct everywhere else.

**Two things worth taking away from this, neither of which is a criticism of the patch:**

Your three new tests are good tests and they pass — but they assert what `parse_proxy_line` *returns*, and this bug lives in what aiohttp *does with it*. No unit test at that boundary could have caught it. That is not a gap you should have filled; it is why the issue needed to be right.

And the deprecation turns out not to be fixable at all right now. `proxy_auth` rejects anything that is not a `BasicAuth` (`raise ValueError("proxy_auth must be None or BasicAuth() tuple")`), and `BasicAuth` is the thing that warns. aiohttp deprecated the only type its own parameter accepts without shipping a replacement.

**So I am not asking you to revise this into something else** — there is nothing correct to revise it into until aiohttp 4 provides a proxy-auth API. I have rewritten #151 to be what it should have been: keep `BasicAuth`, and silence the warning as noise if it is worth silencing.

**If you would like that one instead, it is yours** — it needs a `pytest.ini` (or a `[tool.pytest.ini_options]` block; this repo has neither yet) with a `filterwarnings` entry and a comment pointing at the thread so nobody re-opens this in six months. Small, and it would put your name on the changelog for finding the real answer here rather than the one I guessed at. Say the word and I will assign it.

If you would rather not spend another evening on my mistake, that is completely fair too, and thank you for this one regardless.

LeyckerS:
@slegarraga — closing this, thirteen days on from the review above, because there is nothing left for the branch to target: #151 was rewritten (the original instruction was mine and wrong), the narrow fix landed as #161, and the issue has been closed since 12 August.

This closure is bookkeeping, not a verdict. The patch did what it was asked, quickly and carefully, on a first pull request — and the investigation it triggered is now recorded in three places: the corrected #151, the `pytest.ini` comment that warns the next person, and the review above explaining why header-based proxy auth cannot work for HTTPS. That is real value, even though the diff itself could not ship.

The offer from the review stands in spirit: the roadmap (#39) is live, and if anything there catches your eye, you would be welcome.

PR #153, opened 2026-08-07
feat(cli): structured exit codes for moon_cli.py

### LINKED ISSUE #32: enhancement: give moon_cli.py meaningful exit codes for scripting

AdvaitVarhade:
Fixes #32 by implementing structured, predictable exit codes in `moon_cli.py` to allow automated callers to reliably distinguish success, partial failure, complete failure, and pre-flight configuration errors.

### Changes
- **Exit Code 0**: Every file in the batch completed successfully (`fail_count == 0` and `ok_count > 0`).
- **Exit Code 1**: Run completed but some URLs failed (`fail_count > 0`), run was interrupted (`KeyboardInterrupt`), or an unhandled crash occurred.
- **Exit Code 2**: Pre-flight error where execution could not start (missing/empty `--urls` file).
- **Exit Code 3**: Total failure where every attempted URL failed (`fail_count > 0` and `ok_count == 0`).
- Documented these codes in `docs/CLI.md`.
- Added automated coverage for all exit codes in `tests/test_cli_exit_codes.py`.

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -114,11 +114,12 @@ causes the shared Chrome instance to start; it is not one browser per worker.
 
 ## Exit codes
 
-| Code | Current behavior |
+| Code | Meaning |
 |:--:|:--|
-| `0` | The CLI completes normally, including a run in which individual URLs fail, or is interrupted with `Ctrl+C`. Inspect the final `ok` and `fail` counts and `failed_links.txt` when it is written. |
-| `1` | The URL file is missing or has no usable URLs, or an unhandled exception reaches `main()`. In the latter case the CLI writes `crash_log.txt` beside `moon_cli.py`. |
-| `2` | `argparse` rejects command-line usage, such as missing required arguments or a non-integer value for an integer argument. |
+| `0` | **Success** — Every file in the batch completed successfully. |
+| `1` | **Partial failure / Interrupted** — The run completed but at least one file failed (dead link, exhausted retries, etc.), OR the run aborted completely (e.g. `ENOSPC` disk full) before the batch finished, OR the run was interrupted via `Ctrl+C`, OR an unhandled runtime exception occurred (`crash_log.txt` written). |
+| `2` | **Pre-flight error** — The run could not start due to invalid usage (e.g. `argparse` missing arguments), the `--urls` file not being found, or the file containing no usable URLs. |
+| `3` | **Total failure** — The run completed but every single URL attempted failed. |
 
 After a normal run, the CLI writes `moontech_cli_*.log` and `moontech_cli_*.json`
 beside `moon_cli.py`, not inside `--output`. If any URL exhausts its attempts, it
--- a/moon_cli.py
+++ b/moon_cli.py
@@ -342,6 +342,8 @@ async def run(urls: list[str], output_dir: str, n_workers: int,
         except OSError as e:
             print(f"[warn] Failed links save error: {e}")
 
+    return ok_count, fail_count, disk_full is not None
+
 # ── ENTRY POINT ────────────────────────────────────────────────────────────────
 def main():
     ap = argparse.ArgumentParser(
@@ -368,13 +370,13 @@ def main():
         n_extractors = 8
 
     if not os.path.exists(args.urls):
-        print(f"ERROR: urls file not found: {args.urls}"); sys.exit(1)
+        print(f"ERROR: urls file not found: {args.urls}"); sys.exit(2)
 
     with open(args.urls, encoding="utf-8", errors="replace") as f:
         urls = [l.strip() for l in f if l.strip() and not l.startswith("#")]
 
     if not urls:
-        print("ERROR: no URLs found in file"); sys.exit(1)
+        print("ERROR: no URLs found in file"); sys.exit(2)
 
     print(f"Loaded {len(urls)} URLs from {args.urls}")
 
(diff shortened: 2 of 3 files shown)


LeyckerS (commented):
@AdvaitVarhade — the exit-code work itself is right, including the detail that is easy to get wrong: total failure is tested before partial failure, so code 3 is reachable.

```python
if ok == 0 and fail > 0:
    sys.exit(3)
elif fail > 0:
    sys.exit(1)
else:
    sys.exit(0)
```

Reversed, `exit(3)` would be dead code. Documenting the scheme in `docs/CLI.md` and covering each branch in `tests/test_cli_exit_codes.py` is exactly what #32 asked for.

**Two things before this can go in.**

**1. `moon_extract.py` is outside the scope of #32.** The narrowing in `_extract_datanodes_on_context` is #46 / #130 work — good work, and it belongs in a pull request that says so. You know this rule better than most; you have been on the right side of it several times. The changelog quotes these descriptions, and "structured exit codes for `moon_cli.py`" cannot honestly cover an exception-handling change in the extractor. Please take it out and open it separately — I will review it on its own and it will not be a hard sell.

**2. While it is out, this is worth knowing.** The new handler is:

```python
except (PlaywrightError, PlaywrightTimeoutError):
```

Those are the same set. Verified rather than assumed:

```python
>>> from playwright.async_api import Error as E, TimeoutError as T
>>> issubclass(T, E)
True
>>> [c.__name__ for c in T.__mro__]
['TimeoutError', 'Error', 'Exception', 'BaseException', 'object']
```

`PlaywrightTimeoutError` subclasses `PlaywrightError`, so the tuple catches precisely what `except PlaywrightError:` catches. The behaviour is correct — timeouts *are* caught, which is what you want on a `wait_for_load_state(..., timeout=25000)` — but it reads as though two distinct failures are handled, and a future reader may "simplify" it by removing the wrong half. This is the same shape as the note on #130; naming only `PlaywrightError` and letting the comment say that timeouts are included is clearer. Not blocking, and not blocking in the follow-up PR either — just do not lose the comment.

**Also: this is currently conflicting with `main`.** #150 landed since you opened it and rewrote a large part of `moon_cli.py`, including the `try/except` block at the bottom that your exit codes attach to. The run function's return is part of that region, so the rebase is not purely mechanical — check that `ok`/`fail` still mean what you expect after the disk-full changes, in particular that a run aborted by `ENOSPC` produces the exit code you want. My reading is that it should be 1 rather than 3 even when nothing succeeded, because the run did not fail per-URL — but that is a judgement call and I would rather you make it deliberately and write it into `docs/CLI.md` than inherit it by accident.

#154 also touches `moon_cli.py`. I will sequence the two of you; no coordination needed on your side.

LeyckerS (approved):
`moon_extract.py` is out — thank you. What is left is exactly #32 and nothing else.

Checked the rebase rather than assuming it was mechanical, since #150 rewrote the region this attaches to:

**The `aborted` branch is the right call and it is first.**

```python
if aborted:      sys.exit(1)
elif ok == 0 and fail > 0: sys.exit(3)
elif fail > 0:   sys.exit(1)
else:            sys.exit(0)
```

A run stopped by a full disk reports 1, not 3, even when nothing succeeded — because the URLs did not individually fail, the run was cut short. That is the judgement I would have made and you have written it into `docs/CLI.md`, which is what I care about more than the number itself.

**`run()` still has one exit point.** I checked this specifically, because a three-value unpack in `main()` would raise `TypeError` if any path returned `None`. The two `return`s inside the function body are both at twelve-space indentation inside the nested `do_dl` coroutine; `run()` itself falls through to your new `return ok_count, fail_count, disk_full is not None`. Safe.

**Verified end to end** on your branch rather than from the tests alone:

```
missing urls file  -> 2
empty urls file    -> 2
argparse error     -> 2
```

That last one is worth noting in your favour: argparse already exits 2 on a usage error, so choosing 2 for pre-flight problems makes the CLI consistent with its own parser instead of inventing a third convention. `pytest tests/ -q` → 50 passed.

One behaviour change that is correct but worth flagging in the changelog rather than burying: the setup errors moved from **1 to 2**. Any script that currently tests `if [ $? -ne 0 ]` is unaffected, but one testing `-eq 1` specifically would change meaning. It is the right move — that is the whole point of the issue — and I will call it out in the release notes.

Merging.

Issue #83, opened 2026-08-01
gui: Extractors sits under "COMMON — BOTH METHODS" but datanodes is hard-capped at 8 lanes


LeyckerS:
Found by running the program, not by reading it — the operator set **Extractors 32** and observed that datanodes was far slower than fuckingfast on the same batch, and suspected the "common" settings were not really common. They were right.

## What the code does

`moon_extract.py:878`:

```python
DN_LANES = max(1, min(int(os.environ.get("MOON_DN_LANES", "3") or 3), 8))
```

Datanodes extraction runs through `_lane_queue`, which holds exactly `DN_LANES` permits. The docstring at `moon_extract.py:1229` is explicit: *"Blocks (awaits, does not spin) once DN_LANES pages are already open — excess workers queue here instead of piling more tabs."*

So datanodes extraction is capped at **8 concurrent pages, and only 3 unless raised**. Setting Extractors to 32 gives datanodes no additional concurrency at all: the other 24+ workers simply block on the lane queue.

For fuckingfast — pure HTTP, no lanes — all 32 workers really do run.

## Why this is a bug and not just documentation

The GUI panel is titled **"COMMON · BOTH METHODS"** and Extractors is the first slider in it, with "rec. 16" next to it. That tells the operator the value applies to both providers. It does not. The provider-specific control for datanodes is `Pages` (`dn_pages`), which lives in a different panel.

The measured consequence, same 82-link batch: fuckingfast finished in 4m02s, datanodes took 6m09s for 72 links (10 were genuinely dead). The operator reasonably concluded something was structurally wrong with the settings.

## What needs deciding

Either:

1. **Move Extractors out of the COMMON panel**, or annotate it — something like "datanodes is limited to `Pages` lanes regardless of this value", or
2. **Clamp what the slider offers when it cannot be honoured**, so the number shown is the number that runs, or
3. **Let `dn_pages` follow Extractors** where the 8-lane ceiling allows it, so one control means one thing

Option 1 is the smallest and the least likely to break anything. Whoever takes this should say which they picked and why.

Related to #45, which is the same class of problem on the CLI side: a control whose name promises something it does not do.

## Files

- `web/index.html` — the panel markup and its heading
- `web/app.js` — the slider wiring
- `moon_extract.py:878` — where the ceiling is set

LeyckerS:
**Measured, on two runs of the same 82-link batch.** This is no longer just a reading of the code.

| | fuckingfast | datanodes |
[...]

PR #156, opened 2026-08-07
fix(tests): return actual download count in run_cli to catch missing output files

### LINKED ISSUE #155: test_no_chrome.py reports success when the run produced no files, so the ok assertion cannot fail
Found by @XEDAB while working on #97 — their branch is the first thing that ever made this assertion do work, and it failed immediately. Details and the measurement are in [this thread](https://github.com/LeyckerS/moondownloader/issues/97#issuecomment-5218806383).

## The problem

`tests/test_no_chrome.py:50`:

```python
def run_cli(urls, retries=1) -> dict:
    with tempfile.TemporaryDirectory() as out:
        asyncio.run(moon_cli.run(urls, out, 4, 4, retries, "proxies.txt"))
        done = len(os.listdir(out))
    return {"ok": len(urls) if done == 0 else done, "fail": 0}
```

If the run writes **no** files, `done == 0` and `ok` becomes `len(urls)`. `assert_browser_counts` then asserts:

```python
assert result["ok"] == len(urls)
```

which is `len(urls) == len(urls)`. The zero-file case — the one worth catching — is reported as a complete success.

## It is the case that actually happens

Measured on current `main` with a temporary probe using the same fixtures and the same URLs as `test_cli_mixed_batch_launches_one_shared_browser`:

```
FILES_WRITTEN: 0 []
```

So that assertion has never been able to fail on any run of this suite. The `open_browser` / `playwright` / `close_browser` counts above it are real and do their job; only the `ok` line is inert.

Three tests share `assert_browser_counts` and all three inherit this:

- `test_cli_fuckingfast_only_launches_no_browser`
- `test_cli_mixed_batch_launches_one_shared_browser`
- `test_cli_datanodes_only_launches_one_browser`

## The decision to make

There is a real choice here and I do not want to prescribe the wrong one again:

1. **Make the stub write files**, so `done` is a true count and the assertion means "every URL produced a file". Strongest, but it puts the stubs on the hook for producing believable output and may need the download layer stubbed further down than it currently is.
2. **Drop the `ok` assertion** and replace it with something the stub actually guarantees — the browser counts already carry the intent of these tests, and a fourth assertion that cannot fail is worse than three that can.

Option 2 is smaller and honest. Option 1 is more valuable if the stub can support it without becoming a second implementation of the downloader. Whoever takes this should pick one and say why in the PR body; either is acceptable, an assertion that silently cannot fail is not.

## Done when

- No assertion in `test_no_chrome.py` passes by construction
- If the `ok` check survives in any form, it fails when the run produces fewer files than URLs — demonstrate this, e.g. by breaking something locally and showing the test goes red
- `pytest tests/ -q` green

Needs no Windows machine. The stubs replace Chrome and the network, so this runs anywhere.

snushev:
Closes #155. Removed the fallback that forced ok count to len(urls) when done == 0, preventing false positive test passes when no files are actually written.

### DIFF
--- a/tests/test_no_chrome.py
+++ b/tests/test_no_chrome.py
@@ -3,6 +3,7 @@
 Network, extraction and Chrome are stubbed at the moon_extract level: what is
 under test is the dispatcher's decision to launch, not the extractors.
 """
+
 from __future__ import annotations
 
 import asyncio
@@ -53,7 +54,7 @@ def run_cli(urls, retries=1) -> dict:
     with tempfile.TemporaryDirectory() as out:
         asyncio.run(moon_cli.run(urls, out, 4, 4, retries, "proxies.txt"))
         done = len(os.listdir(out))
-    return {"ok": len(urls) if done == 0 else done, "fail": 0}
+    return {"ok": done, "fail": len(urls) - done}
 
 
 def assert_browser_counts(result, calls, urls, want_browsers: int) -> None:


snushev:
My bad, noticed the mocked behavior

LeyckerS:
@snushev — you spotted it yourself and closed it before anyone had to say a word, which is the right instinct and I would rather have that than a defended patch. No harm done at all.

For the record, since it is a genuinely tricky one: the stubs write no files, so `len(os.listdir(out))` is 0 and `run_cli` substitutes `len(urls)`. Removing the fallback alone leaves `ok == 0` against `len(urls) == 4`, which turns a vacuous pass into a false failure. And it turns out there is a further wrinkle neither of us had: the same helper is used by the **engine** tests, where `ok` comes from `engine.snapshot()` and is real — so the assertion is dead through two callers and live through two others. I had that wrong in the issue too and have corrected it there.

#155 was already reserved for @XEDAB, who reported it — that is on my listings for not making it obvious enough, not on you.

If you would like something that is genuinely free right now, **#83** is the most interesting one open: the interface offers one "streams" number for both providers while datanodes is hard-capped at 8 lanes, so the number on screen is not the number in effect — measured at roughly 40 of 48 slots busy for one provider against 4 for the other. **#151** is smaller and more self-contained if you would rather. Say the word on either and it is yours.

PR #158, opened 2026-08-08
ci: run ruff once outside syntax matrix

### LINKED ISSUE #81: ci: ruff runs three times per pull request, once per Python version, checking the same thing
`.github/workflows/lint.yml` runs the `ruff` steps inside the `syntax-check` job, and #69 put that job on a `[3.10, 3.11, 3.12]` matrix. So `pip install ruff` and `ruff check .` now execute three times on every pull request.

Linting is independent of the interpreter version — `ruff.toml` pins `target-version = "py310"` explicitly, so all three runs check identical rules against identical files and always reach the same verdict. Two of the three are pure cost, and they also make a lint failure show up as three red checks instead of one.

Byte-compiling genuinely does need the matrix: that is the part that can differ between versions, and it should stay where it is.

## The change

Move the two ruff steps out of `syntax-check` into their own job that does not use a matrix. Something like:

```yaml
  ruff:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v7

      - name: Set up Python
        uses: actions/setup-python@v6
        with:
          python-version: "3.10"

      - name: Install ruff
        run: pip install "ruff==0.16.1"

      - name: Lint with ruff
        run: ruff check .
```

Keep the version pin. An unpinned linter turns an unrelated pull request red the day upstream adds a rule.

## Acceptance criteria

- `ruff check .` runs exactly once per pull request
- The `syntax-check` matrix still byte-compiles on 3.10, 3.11 and 3.12
- All jobs green

One file, no Windows machine needed. Worth pairing with #74, which fixes the other problem in the same workflow.

nightcityblade:
## Description

Fixes #81

Move the pinned Ruff install and lint steps into a standalone Python 3.10 job. Ruff now runs once per workflow while `syntax-check` continues to byte-compile every tracked Python module on Python 3.10 through 3.14.

## Type of change

- [ ] Documentation update
- [ ] Refactor / code cleanup
- [x] Other: CI workflow maintenance

## Checklist

      description

Shared extraction and download logic is unaffected; this change only adjusts the lint workflow.

## Testing

- `.venv/bin/ruff check .`
- `python3.10 -m compileall -q`, through `python3.14 -m compileall -q`, on every tracked Python file
- `actionlint .github/workflows/lint.yml`

N/A — CI-only change.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -36,6 +36,17 @@ jobs:
       - name: Byte-compile every module
         run: git ls-files '*.py' | xargs python -m compileall -q
 
+  ruff:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Set up Python
+        uses: actions/setup-python@v7
+        with:
+          python-version: "3.10"
+
       - name: Install ruff
         run: pip install "ruff==0.16.1"
 


LeyckerS (approved):
Clean, and the diff is smarter than its size suggests: `+11 -0` because you inserted the new job header above the existing steps and re-parented them, rather than deleting and retyping. Reads perfectly in the resulting file.

Checked the outcome rather than the diff:

- `syntax-check` keeps the 3.10–3.14 matrix and now byte-compiles only.
- `ruff` is a standalone job on a single interpreter.
- CI reports one `ruff` check against five `syntax-check` entries — previously it was five ruff runs of identical work.

**You kept the pin.** `ruff==0.16.1` is the detail I flagged when you claimed this, and it is the one that matters: an unpinned linter turns someone else's unrelated pull request red on the day upstream adds a rule. Thank you for holding it.

Also correct that ruff runs on 3.10 rather than the newest — the linter's own behaviour should not vary with the interpreter it happens to be installed under, and `target-version = "py310"` in `ruff.toml` is what actually decides what it checks.

`actionlint` in your testing notes is a nice touch; nothing in this repository asks for it.

Merging.

Issue #145, opened 2026-08-05
THEME color block in moon_engine.py: zero references, tests pass without it — but why is it here?


XEDAB:
There's a 15-color theme block at the top of `moon_engine.py` (lines 45–60: `BG`/`ACC`/`GOLD`/`TEXT`… under a `# ── THEME ──` comment).

I grepped the whole repo for these names — bare references, `moon_engine.<name>` attribute access, imports, even the hex values themselves — zero references anywhere. Then I commented the block out and ran the test suite: all 24 tests pass. So removing it breaks nothing.

I'll admit these color blocks are BEAUTIFUL, but I'm curious why they're living in a backend engine module instead of the web front-end. Leftover from something? Could they be cleaned up, or moved somewhere they'd actually be used?

<img width="530" height="535" alt="Image" src="https://github.com/user-attachments/assets/b633b566-7c8f-427d-9e8a-2ea6a2012fb6" />

LeyckerS:
You are right on all three counts, and the answer to "why is it here" is archaeology rather than design.

`moon_engine.py` used to be a **generated file**. The history:

```
43114d5  refactor: gen_1.py -> moon_tk.py, and the engine regenerated from it
f2af536  engine: moon_engine.py is the source now, not a generated file
5f172d3  chore: delete the tkinter GUI and its generator
```

There was a tkinter GUI, `moon_tk.py`, and a `build_engine.py` that produced `moon_engine.py` from it by stripping the interface. That block is **the tkinter colour palette** — `BG`, `SURFACE`, `BORDER`, `TEXT`, `OK`, `ERR` are Tk widget colours, and the generator carried the module-level constants across wholesale because separating "engine" from "interface" by transformation is imprecise at the edges.

Then `moon_engine.py` became the source of truth in its own right, and `moon_tk.py` and the generator were deleted. The palette stayed because by then it was simply part of the file. None of the three exist any more:

```
$ ls moon_tk.py build_engine.py start_tk.bat
  nothing
```

The GUI that replaced it is `web/styles.css`, which has its own palette in CSS custom properties and shares not one hex value with this block.

**So: delete it.** Your check was the right one and it is complete — no bare references, no attribute access, no imports, no hex-value matches, and the suite passes without it. There is no consumer to break because the consumer was deleted in `5f172d3`.

[...]

PR #159, opened 2026-08-08
removed the dead theme block in moon_engine.py

AashishGupta2007:
## Description

 I have removed the dead theme block at the top of moon_engine.py as it had no use in the code due to the introduction of web/styles.css, which has its own palette in CSS custom properties and shares not one hex value with this block.

Addresses #145

## Type of change

- [ ] Documentation update
- [x] Refactor / code cleanup
- [ ] Other:

## Checklist

      — not applicable: this removes an unused constant block with zero
      references, no shared logic touched.
      description

Screenshot 2026-08-08 at 1 49 13 PM

### DIFF
--- a/moon_engine.py
+++ b/moon_engine.py
@@ -44,22 +44,7 @@ from moon_download import (
     RunFatalControl,
 )
 
-# ── THEME ──────────────────────────────────────────────────────────────────────
-BG      = "#080b12"
-BG2     = "#0c1018"
-BG3     = "#111520"
-SURFACE = "#161c2a"
-BORDER  = "#1e2840"
-ACC     = "#00d4ff"
-ACC2    = "#0099cc"
-ACC3    = "#00ffb3"
-GOLD    = "#f5a623"
-TEXT    = "#e8f0ff"
-TEXT2   = "#8899bb"
-TEXT3   = "#3d506e"
-OK      = "#00e676"
-ERR     = "#ff4d6d"
-WARN    = "#ffb547"
+
 # ── EXTRACTION ────────────────────────────────────────────────────────────────
 # Both host front-ends changed in 2026; the extraction layer now lives in
 # moon_extract.py so the engine and the CLI share one implementation.


LeyckerS (approved):
Verified all three things this rests on rather than taking the description's word for it.

**All fifteen names are unreferenced.** Worth showing the working, because a naive `grep -w OK` returns six hits and looks alarming:

```
integration_http.py:214:    print("[OK] browser → HTTP loopback → Engine: ...")
integration_web.py:189:    print("[OK] GUI ↔ bridge ↔ engine: ...")
moon_download.py:310:        W(f"OK: {len(ok_r)}  /  Fail: {len(fail_r)}")
moon_download.py:365:        W(f"  Completed OK   : {len(ok_r)}")
moon_download.py:407:        W(f"  {'Time':>7}  ... {'OK':>5} ...")
render_gui.py:83:        print("[OK] no js errors, no overflow, no horizontal scroll")
```

Every one is inside a string literal. The only real definition is the line you are deleting.

**Your `styles.css` claim holds exactly as stated.** I checked all fifteen hex values against `web/styles.css` and not one appears. That is a stronger check than "it looks unused" and it is the reason this is safe rather than merely plausible.

**On #145's actual question — "why is it here?"** — the answer is in the history, and it is worth recording since the issue asked for it:

```
5f172d3  chore: delete the tkinter GUI and its generator
43114d5  refactor: gen_1.py -> moon_tk.py, and the engine regenerated from it
```

`moon_engine.py` was once **generated** from a tkinter GUI. The palette was that GUI's, compiled into the engine along with everything else, and it survived `5f172d3` because deleting the generator did not touch its output. So this is not a constant someone added and forgot — it is residue from a build step that no longer exists. Module-level constants with zero references in this file are usually from that era, which is a useful thing for the next person to know.

Your reasoning about `web/styles.css` explains why it is *unused today*; the history explains why it *exists at all*. Both halves together are the complete answer, and I will put the second one in the changelog alongside your first.

CI green on all eight. Merging.

PR #161, opened 2026-08-09
test: suppress aiohttp BasicAuth deprecation warning

### LINKED ISSUE #151: silence the aiohttp BasicAuth deprecation warning in the test run (the call itself has to stay)

Divesh-Kshirsagar:
## Description

Adds a targeted pytest warning filter for the `aiohttp.BasicAuth` deprecation warning in `pytest.ini`.

`aiohttp.BasicAuth` is intentionally kept because it is still required for proxy authentication, including HTTPS proxy connections. The warning is suppressed only during pytest runs rather than changing the application code.

## Type of change

* [x] Bug fix (non-breaking change that fixes an issue)
* [ ] New feature (non-breaking change that adds functionality)
* [ ] Breaking change (fix or feature that would change existing behavior)
* [ ] Documentation update
* [ ] Refactor / code cleanup
* [ ] Other:

## Checklist

* [x] I have tested my changes locally
* [ ] If this affects shared logic (extraction, download engine), I also
* [x] I have kept the single-file architecture (no package split)
* [x] I have not added new dependencies without justification in the PR
  description

Tests:

* `uv run pytest tests/` — 50 passed
* `act -j no-chrome` — passed
* `act -j ruff` — passed

Closes #151

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,6 @@
+[pytest]
+filterwarnings =
+    # aiohttp.BasicAuth is deprecated, but proxy_auth still requires it.
+    # Keep BasicAuth because header-based proxy auth fails for HTTPS connections.
+    # See issue #151.
+    ignore:BasicAuth is deprecated and will be removed in aiohttp 4\.0:DeprecationWarning


XEDAB (approved):

LeyckerS (approved):
This is exactly what #151 asked for, and it gets the part right that the issue was actually about.

Verified both "Done when" criteria locally rather than reading the diff:

**The target warning is gone.** `pytest tests/ -q` → 50 passed, with no warnings summary at all; it previously ended `50 passed, 1 warning`.

**An unrelated deprecation still shows.** I added a throwaway test raising a different `DeprecationWarning`:

```
tests/test_zz_warn.py::test_other_deprecation_still_visible
  DeprecationWarning: a completely unrelated deprecation
1 passed, 1 warning
```

Still reported. That is the difference between this and a blanket `ignore::DeprecationWarning`, and it is the whole reason the issue was written the way it was — matching on the message text rather than the category is what keeps the next real deprecation visible.

The comment is right too, and it is the actual deliverable here. `proxy_auth` rejects anything that is not a `BasicAuth`, and moving the credentials into a header breaks HTTPS proxies specifically, because aiohttp authenticates those on the `CONNECT` request rather than inside the tunnel. Someone will find that suppression in a year and "fix it properly"; your three lines are what stops them. Pointing at #151 means the full reasoning is one click away.

`moon_download.py` untouched, as required.

One thing I noticed that is **not** yours to fix and that I am not asking for: neither workflow watches `pytest.ini`, so a future change to this file will not trigger CI at all. #151 said nothing about path filters and I am not moving the goalposts after the fact — opening it separately.

Merging. Welcome, and thank you for reading the issue closely enough to keep the narrow match.

PR #162, opened 2026-08-09
fix: apply the download_file stub to moon_engine too (#160)

### LINKED ISSUE #160: conftest stubs download_file on moon_cli only — the engine keeps the real one, and nothing currently notices
Found while verifying #157. Not currently causing a failure, which is exactly why it is worth writing down.

## The problem

`tests/conftest.py`, lines 61–65:

```python
for host in (moon_engine, moon_cli):
    monkeypatch.setattr(host, "extract_fuckingfast", fake_ff)
    monkeypatch.setattr(host, "extract_datanodes", fake_dn)
    monkeypatch.setattr(host, "close_ff_session", lambda: asyncio.sleep(0))
monkeypatch.setattr(moon_cli, "download_file", fake_download)
```

The extractors are stubbed on **both** front-ends. `download_file` is stubbed on **`moon_cli` only** — that last line sits outside the loop. `moon_engine` imports `download_file` directly from `moon_download`, so the engine tests hold the real function.

## Why nothing has failed

`run_engine` in `tests/test_no_chrome.py` starts the engine with `"mode": "links"`, which extracts and never downloads. Measured with a temporary probe:

```
PROBE ok/fail/dl_done/dl_total: 4 0 0 4
PROBE files in _tmp_out: []
```

Four successes, zero downloads, nothing on disk. The real `download_file` is present but never called, so the gap is invisible.

## Why it matters anyway

The suite's whole promise is that it needs neither a browser nor a network — `no-chrome` runs on a bare CI box with only `aiohttp`, `curl_cffi` and `pytest` installed. That promise currently holds by accident on the engine side: change `run_engine` to `"mode": "download"`, or add any engine test that downloads, and the suite starts making real HTTPS requests to `dl.fuckingfast.co/...?fake` from CI. It would be slow, flaky, and confusing to diagnose, and the person who trips it will not be looking at `conftest.py`.

It also has a live consequence worth knowing: `result["ok"]` from `run_engine` counts **extracted links** (`moon_engine.py:270`), not completed downloads. Anyone reading those tests would reasonably assume otherwise — I did.

## The work

Move the `download_file` stub inside the loop so both front-ends get it, then confirm nothing depended on the asymmetry:

- `pytest tests/ -q` still green
- The engine tests still fail when the engine drops a URL. Demonstrate it — dropping the `self._inc("_ok")` at `moon_engine.py:270` should turn `test_engine_fuckingfast_only_launches_no_browser` and `test_engine_mixed_batch_launches_one_shared_browser` red. That is how #157 was verified and it takes a minute.

If moving it turns something red, **say so on the issue rather than working around it** — a stub that cannot be applied to the engine would be telling us something about the engine, and I would want to see that before it gets patched over.

Optional, and worth more than the fix: add a short comment above the loop saying that both front-ends must be stubbed and why, so the next edit does not re-introduce the split.

Small. No Windows machine needed.

XEDAB:
## Description
Close #160
Resolved the issue in #160 by modifying the erroneous line 66 in conftest.py (monkeypatch.setattr(moon_cli, "download_file", fake_download))) (which could potentially cause the engine to attempt a real download) to monkeypatch.setattr(host, "download_file", fake_download)
图片 and moved it into the for host in (moon_engine, moon_cli): loop, while adding a comment stating that this monkeypatch statement should not be moved.
This modification only involves conftest.py and does not involve any other changes.
## Type of change

- [ ] Documentation update
- [ ] Refactor / code cleanup
- [ ] Other:

## Checklist

      description

图片
↑Normal state
图片
↑The state after commenting out lines 181 and 255

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -59,9 +59,10 @@ def browser_calls(monkeypatch):
     monkeypatch.setattr(moon_extract, "shutdown_chrome", fake_shutdown_chrome)
 
     for host in (moon_engine, moon_cli):
+        # Every stub in this loop must cover both moon_engine and moon_cli. Break that and the engine keeps the real function — it will actually try to make network requests, because monkeypatch never replaced its download_file.
         monkeypatch.setattr(host, "extract_fuckingfast", fake_ff)
         monkeypatch.setattr(host, "extract_datanodes", fake_dn)
         monkeypatch.setattr(host, "close_ff_session", lambda: asyncio.sleep(0))
-    monkeypatch.setattr(moon_cli, "download_file", fake_download)
+        monkeypatch.setattr(host, "download_file", fake_download)
 
     return calls


XEDAB:
By the way, I looked at line 270 mentioned in #160; it is a comment. I commented out the two lines of self._inc("ok") (181 and 255) in the fuckingfast branch, and the test crashed as expected.

LeyckerS (approved):
Correct, and verified rather than assumed:

```
ENGINE download_file: fake_download
CLI    download_file: fake_download
```

Both front-ends now hold the stub; before this, the engine held the real function. `pytest tests/ -q` → 50 passed.

Your verification in #160 is the part I want to point at. You did not just move the line — you went to `moon_engine.py:270`, found it was a comment rather than the counter, located the real increments at 181 and 255, annotated those instead, confirmed the tests went red, and restored. **The line number in my issue was wrong and you worked past it without making a fuss.** That is the third time you have had to correct something I wrote, and you have been right every time.

Putting the stub inside the loop rather than adding a second `setattr` beside it is also the right shape: the bug was that one line sat outside the loop, so the fix is to make the loop the only place any of this happens. Your comment says exactly that, which is what stops it recurring.

One cosmetic thing I am not sending you back for: that comment is a single 228-character line, against `line-length = 100` in `ruff.toml`. It only passes because `E501` sits in the ignore list, and every other comment in the file wraps. I will wrap it in the bookkeeping commit rather than cost you a round trip — the wording is yours and I am not changing it.

Merging.

Issue #79, opened 2026-08-01
ruff.toml: the E501 comment says line length is handled elsewhere, and it is not


LeyckerS:
**This issue is reserved for someone who has never had a pull request merged on GitHub before.** If you have contributed before, please leave it — [the roadmap](https://github.com/LeyckerS/moondownloader/issues/39) has plenty of other work.

There is no programming here. You are changing one line of a comment. The point is to walk you through the process once.

## The problem

`ruff.toml` sets a maximum line length and then switches the check for it off. The comment claims the length is "handled above via line-length instead" — but the `line-length` setting is the value that check reads. Turning the check off does not move the enforcement somewhere else; it removes it. The file tells the next reader something untrue.

## The change

**File:** `ruff.toml` — **line 6**

Currently:

```toml
    "E501",  # line length - handled above via line-length instead
```

Change it to:

```toml
    "E501",  # line length is not enforced: ruff format is not run in CI
```

That is the whole change. One line.

## How to do it, step by step

1. Click **Fork** at the top right of [this repository](https://github.com/LeyckerS/moondownloader). This gives you your own copy.
2. In your copy, click on `ruff.toml` to open it, then click the **pencil icon** to edit.
3. Find line 6 and replace it with the new text above.
4. Scroll down. Under **Commit changes**, write a short message such as `docs(ruff): correct the E501 comment` and click **Propose changes**.
5. On the next screen click **Create pull request**, write `Closes #79` in the description, and submit it.

You never have to install anything or run anything. Steps 1 to 5 all happen in the browser.

## Before you start

Comment here saying you are taking it, so nobody else starts the same change. Then go ahead — you do not need to wait for a reply.

If you get stuck at any step, say so in a comment. That is completely fine and it is the point of this issue.

FlaggedATX:
Hi thanks for the opportunity, im taking this

LeyckerS:
@FlaggedATX — yours, assigned, and welcome.

The pull request was already up by the time I read the claim, so the only thing left was to check it: #171 is exactly what this issue prescribed, and the claim the new comment makes is true — I verified it with the CI's own `ruff==0.16.1` pin rather than taking the issue's word for it (details on the pull request). Merging it today.

One small thing for next time, costing you nothing now: the pull request body needs the literal line `Closes #79` for GitHub to close this issue automatically — step 5 above mentioned it, and it is the single most-forgotten step there is. The merge commit will carry it instead, so everything still links up.

Congratulations on the first one.

PR #165, opened 2026-08-12
ci: watch lint and test configuration

### LINKED ISSUE #164: ci: the two files that control what CI checks are the two files CI does not watch
Noticed while merging #161, which added `pytest.ini`. That pull request arrived with **no checks at all** — not pending, not skipped, simply absent — and it was correct behaviour.

## The gap

`.github/workflows/lint.yml` triggers on:

```yaml
paths:
  - "**.py"
  - "requirements.txt"
  - "constraints.txt"
  - ".github/dependabot.yml"
  - ".github/workflows/lint.yml"
```

`docs-cli-check.yml` triggers on `**.md`, `**.py` and itself.

Two configuration files match neither list:

| File | What it controls | Runs it |
|:--|:--|:--|
| `ruff.toml` | line length, and which lint rules are ignored | the `ruff` job |
| `pytest.ini` | which warnings are suppressed during the suite | the `no-chrome` job |

So a pull request that changes only `ruff.toml` does not run `ruff`, and one that changes only `pytest.ini` does not run the tests. **The configuration that decides what CI enforces is invisible to CI.**

## Why this is worth fixing rather than shrugging at

Both files have a plausible bad edit that would sail through with a green tick and no output at all:

- Adding a rule to the `ignore` list in `ruff.toml` silently stops the linter checking it.
- Broadening `pytest.ini`'s filter from the exact `BasicAuth` message to `ignore::DeprecationWarning` would hide every future deprecation in every dependency — which is precisely what #151 was written to avoid, and the comment there says so.

Neither would be caught by review alone, because a reviewer looking at a one-line config diff with all checks green has no reason to suspect anything ran differently.

There is a live example already open: **#79** asks for a one-line correction to a comment in `ruff.toml`. Whoever fixes it will get no CI run on their pull request.

## The work

Add the two filenames to the `paths` lists in `lint.yml` — `push` and `pull_request` both, they are separate lists — so that changing either file runs the job it configures.

Worth thinking about for a moment rather than pasting: `docs-cli-check.yml` does not need them, and adding filters where they are not needed is its own kind of noise. Say in the pull request why you did or did not touch the second workflow.

## Done when

- A pull request touching only `ruff.toml` runs the `ruff` job
- A pull request touching only `pytest.ini` runs `no-chrome`
- **Demonstrate it.** Push a trivial change to one of them on your branch — a comment will do — and link the run that fired. This issue is about CI triggering, so an assertion that it now triggers is worth nothing without the run behind it.
- `pytest tests/ -q` still green

Small, and no Windows machine needed. `#79` and `#93` are reserved for first-time contributors; this one is not, so it is fair game for anyone.

nightcityblade:
## Description

Closes #164.

Adds `ruff.toml` and `pytest.ini` to both the push and pull-request path filters for the lint workflow. This ensures configuration-only changes run the checks they control. `docs-cli-check.yml` is intentionally unchanged because neither file configures documentation checks.

Verification:

- `pytest tests/ -q` — 50 passed
- `uvx ruff==0.16.1 check .` — passed
- Parsed `.github/workflows/lint.yml` with PyYAML — valid
- CI trigger proof: [run 31559809262](https://github.com/LeyckerS/moondownloader/actions/runs/31559809262), fired for the initial commit's temporary comment-only `ruff.toml` change

## Type of change

- [ ] Documentation update
- [ ] Refactor / code cleanup
- [ ] Other:

## Checklist

      applied the equivalent change to `moon_cli.py` (N/A; CI-only change)
      description

N/A — CI-only change.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,6 +7,8 @@ on:
       - "**.py"
       - "requirements.txt"
       - "constraints.txt"
+      - "ruff.toml"
+      - "pytest.ini"
       - ".github/dependabot.yml"
       - ".github/workflows/lint.yml"
   pull_request:
@@ -14,6 +16,8 @@ on:
       - "**.py"
       - "requirements.txt"
       - "constraints.txt"
+      - "ruff.toml"
+      - "pytest.ini"
       - ".github/dependabot.yml"
       - ".github/workflows/lint.yml"
 


LeyckerS (approved):
Correct, minimal, and you answered the question I actually asked rather than the one the diff implies — `docs-cli-check.yml` deliberately untouched, with the reason stated. That was the part I wanted thought about.

**One note on the trigger proof, because it does not prove what it says.** Run [31559809262](https://github.com/LeyckerS/moondownloader/actions/runs/31559809262) is real and green, and its head commit `53b3a21` does touch `ruff.toml` — but it also touches `.github/workflows/lint.yml`, which was **already** in the paths list before this pull request:

```yaml
      - ".github/dependabot.yml"
      - ".github/workflows/lint.yml"
```

So that run would have fired with or without your change. The variable is not isolated. This is not a criticism of the fix — adding two filenames to two lists is correct by inspection and CI is green on all eight checks — it is that the demonstration cannot distinguish the new filter from the old one.

Isolating it would need a commit touching **only** `ruff.toml`, which is awkward to do inside a pull request that by definition also changes `lint.yml`. That is my fault for writing an acceptance criterion that is hard to satisfy in the same branch as the change, so **I will verify it myself on `main` after merging** and post the result here.

Merging. Second one of these from you, and both have been clean.

LeyckerS:
Verified on `main`, as promised.

Commit [`3194603`](https://github.com/LeyckerS/moondownloader/commit/3194603c276b1054c5796a346ff181973d1e6c4f) touches **`ruff.toml` and nothing else** — I checked the commit's file list rather than trusting my own staging:

```
files changed: ruff.toml
```

Result:

```
Lint: completed/success   ← https://github.com/LeyckerS/moondownloader/actions/runs/31641873825
```

**`Lint` fired, and only `Lint`.** `Docs CLI Check` correctly stayed out of it, since it watches `**.md` and `**.py` and neither matched. That is the isolated proof your branch could not produce, and it confirms the filter does what the diff says.

So the practical effect is now real: #79 asks for a one-line correction to a comment in `ruff.toml`, and whoever takes it will get a CI run on their pull request instead of a blank check list. That was the reason the issue existed.

Thanks for this one.

PR #167, opened 2026-08-14
cleanup: drop the unused _LOG_MAX_LINES constant

mazi-eth:
Removed the unused _LOG_MAX_LINES constant and the blank line after it, per the issue.

Ran `grep -rn "_LOG_MAX_LINES" .` after the change — zero matches,
confirming nothing else in the codebase referenced it

### DIFF
--- a/moon_engine.py
+++ b/moon_engine.py
@@ -137,8 +137,6 @@ class Engine:
     def _get(self, attr):
         with self._lock: return getattr(self, attr)
 
-    _LOG_MAX_LINES = 2000
-
     def log(self, msg, tag=""):
         """Thread-safe: called from the asyncio worker, drained by snapshot()."""
         with self._log_lock:


LeyckerS (approved):
Exactly right, and exactly two lines. Verified before merging rather than taking the issue's word for it:

```
$ grep -rn "_LOG_MAX_LINES" --include=*.py --include=*.md --include=*.js .
./moon_engine.py:140:    _LOG_MAX_LINES = 2000
```

One occurrence, which is its own definition. Nothing reads it — the log ring is bounded by `collections.deque(maxlen=6000)` instead, and has been since long before this. All nine checks green.

You also took the blank line with it, which is the small thing that makes a deletion read as finished rather than as a hole left behind.

**This is your first merged pull request here.** Nothing about it needed a second round, which is not the usual experience for a first one anywhere. If you want another, #79 is the remaining reserved issue — a one-line comment in `ruff.toml` that describes its own effect incorrectly — and it now runs CI properly, which it did not a few days ago.

Merging. Thank you, and welcome.

PR #168, opened 2026-08-16
make docs-cli-check watch its own file on PRs

### LINKED ISSUE #93: ci: docs-cli-check.yml watches itself on push but not on pull requests
**This issue is reserved for someone who has never had a pull request merged on GitHub before.** If you have contributed before, please leave it — [the roadmap](https://github.com/LeyckerS/moondownloader/issues/39) has plenty of other work.

You are adding one line. The point is to walk you through the process once, and the reason behind it is worth understanding.

## The problem

`.github/workflows/docs-cli-check.yml` lists the paths that make it run:

```yaml
  push:
    branches: [main]
    paths:
      - "**.md"
      - "**.py"
      - ".github/workflows/docs-cli-check.yml"   # <- it is here
  pull_request:
    paths:
      - "**.md"
      - "**.py"                                   # <- but not here
```

The `push` trigger watches the workflow file itself. The `pull_request` trigger does not.

So a pull request that changes **only this workflow** runs no checks at all. Whoever opens it cannot see whether their change works until after it has already been merged to `main` — which is the worst possible moment to find out.

This is not hypothetical: it happened this week in `lint.yml`. Two people submitted the same fix, and the one whose pull request also added the self-reference came back green while the other ran zero checks. That was #77.

## The change

**File:** `.github/workflows/docs-cli-check.yml`

Add one line under the `pull_request` `paths:` list, so it matches the `push` list:

```yaml
  pull_request:
    paths:
      - "**.md"
      - "**.py"
      - ".github/workflows/docs-cli-check.yml"
```

Indentation matters in YAML — line it up exactly with the two lines above it.

## How you will know it worked

Your own pull request is the test. Because you are changing this file, **and your change makes the workflow watch this file**, the `Docs CLI Check` job should appear on your pull request and pass. If no checks appear, the line is not doing what it should — say so in a comment.

## How to do it, step by step

1. Click **Fork** at the top right of [this repository](https://github.com/LeyckerS/moondownloader).
2. In your copy, open `.github/workflows/docs-cli-check.yml` and click the **pencil icon**.
3. Add the line under `pull_request:` → `paths:`.
4. Under **Commit changes**, write something like `ci: make docs-cli-check watch its own file on PRs`, then **Propose changes**.
5. Click **Create pull request**, write `Closes #93`, and submit.

All of it happens in the browser. Nothing to install.

## Before you start

Comment here saying you are taking it. Then go ahead — no need to wait for a reply.

StefStrg:
closes #93

### DIFF
--- a/.github/workflows/docs-cli-check.yml
+++ b/.github/workflows/docs-cli-check.yml
@@ -11,6 +11,7 @@ on:
     paths:
       - "**.md"
       - "**.py"
+      - ".github/workflows/docs-cli-check.yml"
 
 jobs:
   docs-cli-flags:


LeyckerS (approved):
Correct, and in the right place — which is the only way this one can be got wrong. Checked out the branch to see where the line actually lands:

```yaml
  push:
    paths:
      - "**.md"
      - "**.py"
      - ".github/workflows/docs-cli-check.yml"
  pull_request:
    paths:
      - "**.md"
      - "**.py"
      - ".github/workflows/docs-cli-check.yml"    ← yours
```

`push` already had it; the `pull_request` list did not, so a pull request editing this workflow ran nothing. Now it does. That the check list on this pull request shows only `docs-cli-flags` is itself correct: `lint.yml` does not watch this file, and should not.

The same class of gap was fixed for `ruff.toml` and `pytest.ini` a few days ago in #165, so with this one the configuration files that control CI are all watched by the jobs they configure.

**This is your first merged pull request anywhere**, going by your account history — a good one to start with, because it is small enough to verify completely and it fixes something real rather than cosmetic.

Merging. Welcome, and if you want another, #79 is the last reserved issue.

PR #169, opened 2026-08-17
fix(gui): expose the datanodes extractor limit

### LINKED ISSUE #83: gui: Extractors sits under "COMMON — BOTH METHODS" but datanodes is hard-capped at 8 lanes

yhuikzdtguioaert:
Closes #83

## Description

The shared `Extractors` slider now states the effective datanodes concurrency limit instead of implying that every selected worker can open a datanodes page.

The note is derived from `min(Extractors, Pages)`, updates immediately when either slider moves, and is relabelled when the UI switches between English and Italian. I chose the issue's option 1 because the eight-page ceiling protects the shared Chrome context; explaining the effective limit fixes the misleading control without weakening that guardrail.

## Type of change

- [x] New feature (non-breaking change that adds functionality)
- [ ] Documentation update
- [ ] Refactor / code cleanup
- [ ] Other:

## Checklist

- [x] If this affects shared logic (extraction, download engine), I also applied the equivalent change to `moon_cli.py` (not applicable: presentation-only GUI change)

## Verification

- Regression test against the unmodified tree: `3 failed`
- `python -m pytest tests/ -q`: `53 passed`
- `ruff check .`: passed
- `node --check web/app.js`: passed
- Headless Chromium runtime checks: effective limit changed `8 -> 6 -> 4`, EN/IT relabelling passed, no page errors
- `python render_gui.py <out> 1440 900`: no JS errors, overflow, or horizontal scroll

The extraction and download paths are unchanged; both provider configurations continue to flow through the same existing engine settings.

### DIFF
--- a/tests/test_gui_extractor_scope.py
+++ b/tests/test_gui_extractor_scope.py
@@ -1,0 +1,26 @@
+"""Regression coverage for the provider-specific extractor ceiling in the GUI."""
+
+from pathlib import Path
+
+
+ROOT = Path(__file__).resolve().parents[1]
+INDEX = (ROOT / "web" / "index.html").read_text(encoding="utf-8")
+APP = (ROOT / "web" / "app.js").read_text(encoding="utf-8")
+
+
+def test_extractor_hint_has_a_dedicated_live_target():
+    assert 'id="extractorScope"' in INDEX
+    assert 'id="workers"' in INDEX
+    assert 'id="pages"' in INDEX
+
+
+def test_extractor_hint_reports_the_effective_datanodes_limit():
+    assert "Math.min(workerCount, pageCount)" in APP
+    assert 'T("extractors_hint", Math.min(workerCount, pageCount), pageCount)' in APP
+
+
+def test_extractor_hint_tracks_both_controls_and_language_changes():
+    assert '$("#workers").addEventListener("input", syncExtractorScope)' in APP
+    assert '$("#pages").addEventListener("input", syncExtractorScope)' in APP
+    assert APP.count("syncExtractorScope();") >= 2
+    assert APP.count("extractors_hint:") == 2
--- a/web/app.js
+++ b/web/app.js
@@ -34,6 +34,7 @@ const I18N = {
     mode_download: "Download", mode_links: "Links only",
     common: "Common", common_sub: "· both methods",
     extractors: "Extractors", dl_streams: "DL streams", retries: "Retries",
+    extractors_hint: (n, pages) => `datanodes uses up to ${n} extractors (Pages ${pages})`,
     rec16: "rec. 16", rec8: "rec. 8", rec8p: "rec. 8",
     note_streams: "fewer streams = more bandwidth per file; the pipe is still the ceiling",
     dn_note: "Real Chrome + Turnstile: it opens pages, not windows",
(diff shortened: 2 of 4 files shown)


LeyckerS (approved):
Correct, and you picked the option I would have picked — explaining the effective limit rather than raising the cap. The eight-page ceiling exists because several identities from one IP read as a bot farm to Turnstile, so widening it would have traded a confusing label for a worse failure.

Verified the number is actually right rather than plausible: `dn_pages` is clamped to `(1, 8)` in `Engine._CLAMP` and passed straight to `_moon_extract.configure(lanes=eff["dn_pages"])`, so the datanodes concurrency really is `min(Extractors, Pages)`. The hint says what the engine does.

The details that make this more than a label: it re-runs on `input` for **both** sliders rather than only the one it sits under, it is called from `applyLang()` so switching to Italian relabels it instead of leaving a stale English string, and it degrades safely — `if (!hint || !workers || !pages) return;` means the demo page cannot throw on it. `9/9` checks green, `pytest tests/ -q` passes locally.

**One observation, not blocking.** `tests/test_gui_extractor_scope.py` asserts on the *text* of `app.js`:

```python
assert 'T("extractors_hint", Math.min(workerCount, pageCount), pageCount)' in APP
```

That will fail if someone extracts the expression into a variable, even though the behaviour would be identical — and it would pass if someone kept the string while breaking the wiring around it. Reading source files in tests has precedent here (`test_docs_cli_flags.py` does it), but that one checks documented flags against the *real parser*, so it tests behaviour rather than spelling.

I am not asking you to change it: there is no JavaScript test runner in this repo, you said in the description that it is a UI-contract test, and it does catch the realistic regression, which is the wiring being deleted. Worth knowing what it will and will not catch, and worth loosening if it ever fails for a rename.

Merging. Welcome — a first contribution here that arrived with the reasoning for the design choice already written down is not the common case.

PR #171, opened 2026-08-19
docs(ruff): correct the E501 comment

FlaggedATX:
docs(ruff): correct the E501 comment

## Description

This is a documentation change, replaces the incorrect documentation of E501 (line 10) on ruff.toml for a correct one

## Type of change

- [x] Documentation update
- [ ] Refactor / code cleanup
- [ ] Other:

## Checklist

      description

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -7,7 +7,7 @@ target-version = "py310"
 [lint]
 select = ["E", "F", "W", "I"]
 ignore = [
-    "E501",  # line length - handled above via line-length instead
+    "E501",  # line length is not enforced: ruff format is not run in CI
     # Everything below fires on the codebase as it stands today. Per the issue that
     # introduced this file: the goal is a green baseline that catches regressions, not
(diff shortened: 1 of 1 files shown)


LeyckerS (approved):
Verified rather than read, per the house rule:

- **The diff is exactly the text #79 prescribes** — one line, one file, nothing else touched.
- **The new comment is true.** With this repo's config and the CI's own pin (`ruff==0.16.1`), a 206-character line passes `ruff check` without a murmur, and `.github/workflows/lint.yml` runs `ruff check .` only — `ruff format` appears nowhere in CI. So nothing enforces line length, which is what the line now says. The old comment pointed at `line-length = 100` as if something read it; with `E501` off, nothing does.
- `ruff check .` on the full tree passes, matching the green ruff job here.

Merging. Thank you, @FlaggedATX — a first pull request that says something true about the file it edits is exactly the point of #79.

PR #175, opened 2026-08-22
fix(download): stop counting the final write buffer twice in bytes_acc

### LINKED ISSUE #172: bug: download_file counts the final write buffer twice in bytes_acc — a small file's bytes count double
`bytes_acc` is the deque both front-ends use for the run's byte accounting: the engine sums it for `bytes_total` — the Downloaded card, the end-of-run summary, and the ETA's average-file-size term — and the CLI sums it for the total-GB progress line and the final summary. Every entry is supposed to be bytes received, once.

## The problem

Every chunk is appended as it arrives, `moon_download.py:593`:

```python
bytes_acc.append((now, len(chunk)))
```

The mid-loop flush (`moon_download.py:596-604`) writes the buffer without touching `bytes_acc` — correct. But the tail flush re-appends the whole leftover buffer, `moon_download.py:641`:

```python
if buf:
    data = b"".join(buf)
    ...
    bytes_acc.append((time.monotonic(), len(data)))
```

`data` is the join of chunks that were each already appended at 593. With `WRITE_BUF = 16 MiB` (`moon_download.py:30`), any file smaller than 16 MiB never reaches the mid-loop flush, so **its entire size is counted twice**; a larger file over-counts by its final partial buffer.

## Measured, not argued

Throwaway probe (fake session at the `moon_download` boundary, the `test_on_event.py` pattern), on current `main`:

```
PROBE small.bin: on_disk=5 MiB   bytes_acc_sum=10 MiB   samples=6
PROBE large.bin: on_disk=20 MiB  bytes_acc_sum=24 MiB   samples=21
```

A 5 MiB file records 10 MiB — five chunk entries plus one 5 MiB tail entry. A 20 MiB file records 24 MiB.

## What it distorts

- **`bytes_total` in `Engine.snapshot()`** — on a batch of small files the Downloaded card and the summary read up to **double** the real figure.
- **The engine's ETA input**: `avg_file = total_downloaded / dl_done` (`moon_engine.py:714`) inherits the inflation. #166 is reworking that expression; this bug feeds it wrong numbers from below, and fixing it makes #166's output truer at no cost to either patch.
- **The CLI's totals** (`moon_cli.py:128, 334`) — same deque, same inflation.
- Minor: the tail entry lands inside the 3-second live-speed window at the moment a file completes, so completions produce a brief phantom speed spike in both front-ends.

Not affected: the per-transfer windows `speed_win`/`pub_win` (`moon_download.py:591-592`) are separate deques appended once, so per-file live and average speeds are correct; and nothing about what lands on disk is wrong.

## The fix — verified before writing this

Delete the `bytes_acc.append` at `moon_download.py:641`. The write itself stays; only the double accounting goes. I ran the probe against exactly that one-line change: 5 MiB records 5 MiB, 20 MiB records 20 MiB, on-disk sizes intact, full suite green. Please add a regression test asserting `sum(b for _, b in bytes_acc)` equals the file size for a completed download below `WRITE_BUF` — the probe above is most of that test already, and without one this slips back in silently, which is how it arrived: the re-append came in with the write-buffering in #150 (`a56c6c4`), an accounting slip inside an otherwise careful change, invisible because nothing asserts on the deque's sum.

Size: one deleted line plus a ~50-line test. No Windows needed. To claim it, comment here.

snowyukitty:
Closes #172.

## The defect

`download_file` appends every chunk to `bytes_acc` as it arrives
(`moon_download.py:593`), and the tail flush then re-appended the joined
leftover buffer — bytes that were already counted:

```python
if buf:
    data = b"".join(buf)
    ...
    bytes_acc.append((time.monotonic(), len(data)))   # removed
```

The mid-loop flush at `moon_download.py:596-604` already did the right thing by
leaving the deque alone. With `WRITE_BUF` at 16 MiB, any transfer smaller than
that never reaches the mid-loop flush, so its entire size was recorded twice; a
larger transfer over-counted by whatever remained in the final partial buffer.

## Reproduced on `7e48d31`

Using the repository's own fake-session pattern from `test_on_event.py`:

| transfer | on disk | recorded in `bytes_acc` |
| --- | ---: | ---: |
| 5 MiB | 5,242,880 | 10,485,760 |
| 20 MiB | 20,971,520 | 25,165,824 |

Which matches the figures in the issue.

## The change

One line removed. The write itself, `last_write_bytes` and
`write_persisted_before` all stay; only the duplicate accounting entry goes.

`downloaded`, `rec.done_bytes`, `rec.live_mbs`, `speed_win` and `pub_win` are
untouched — they were already correct, because they are accumulated per chunk
and never from the flush.

## Tests

`tests/test_bytes_acc_accounting.py`, following the fakes in `test_on_event.py`
and `test_resume_200.py`. The cases that probe a buffer boundary derive their
sizes from `WRITE_BUF` and `RECV_CHUNK` so they keep their meaning if those
change; the rest use plain megabyte transfers. Accounting expectations come
from the chunks the fake session actually handed over.

On this branch: 9 passed. On unmodified `main`: 6 fail and 3 pass.

The six that fail are the ones that see the duplicate entry — the two
whole-transfer cases, the three parametrised small transfers, and
`test_one_entry_per_received_chunk`, which catches it as an extra entry rather
than as an inflated total. A representative failure is
`assert 16777216 == 8388608`: an 8 MiB transfer recorded as 16 MiB, with the
deque showing four real chunk entries followed by one 8 MiB tail entry.

The three that pass on `main` do so by design, and they are what keeps the fix
honest rather than just proving the symptom:

- `test_exact_multiple_of_write_buffer_leaves_no_tail` — an exact multiple of
  `WRITE_BUF` leaves `buf` empty, so the tail flush never runs. It pins the
  per-chunk append: a change that moved accounting to the flush sites instead
  would break this case and not the others.
- `test_entry_timestamps_stay_within_the_transfer` — a bound rather than a
  per-chunk claim: it rejects an entry stamped outside the call. Both front-ends
  window this deque at three seconds for the live speed reading, so an entry
  arriving at the wrong time is wrong even when the total is right.
- `test_aborted_transfer_records_only_what_it_received` — a run cut short must
  record what it actually got and nothing more, from either append site. The
  fake records whether the fatal control was already set as it handed each
  chunk over, so the expected entries are the chunks delivered while it was
  clear — an observation rather than an assumption about how far the loop runs
  past the signal.

`ruff check .` is clean and every tracked module byte-compiles.

`tests/test_elapsed_clock.py::test_elapsed_resets_on_new_run` fails on this
machine both before and after this change — it is about the engine's elapsed
clock, not byte accounting, and I have left it alone.

## Not in scope

The ETA expression at `moon_engine.py:714` reads this deque and inherits the
inflation, but that is #166's scope and has an open pull request, so this change
stays out of it.

### DIFF
--- a/moon_download.py
+++ b/moon_download.py
@@ -638,7 +638,6 @@ async def download_file(
                         data = b"".join(buf)
                         if fatal_control.is_set():
                             return False, "aborted_disk_full", downloaded
-                        bytes_acc.append((time.monotonic(), len(data)))
                         last_write_bytes = len(data)
                         write_persisted_before = os.path.getsize(tmp)
                         await loop.run_in_executor(_POOL, _write, f, data)
--- a/tests/test_bytes_acc_accounting.py
+++ b/tests/test_bytes_acc_accounting.py
@@ -1,0 +1,289 @@
+"""Test: every byte download_file receives is recorded in bytes_acc exactly once.
+
+Issue #172. `bytes_acc` is the deque both front-ends read: the engine sums it
+for `bytes_total` and the ETA's average-file-size term, and the CLI sums it for
+the total-GB line and the final summary. Each entry is supposed to be bytes
+received, once.
+
+Every chunk was appended as it arrived, and then the tail flush re-appended the
+whole leftover buffer — bytes that were already counted. With `WRITE_BUF` at
+16 MiB, any file smaller than that never reaches the mid-loop flush, so its
+entire size was counted twice.
+
+The cases that probe a buffer boundary — the multi-buffer transfer, the exact
+multiple, and the small-transfer sweep — derive their sizes from `WRITE_BUF` and
+`RECV_CHUNK` so they keep their meaning if those change. The rest use plain
+megabyte transfers, which is enough for what they assert. Accounting
+expectations come from the chunks the fake session actually handed over. Fakes follow the pattern in
+`test_on_event.py` and `test_resume_200.py`.
+"""
+from __future__ import annotations
+
+import asyncio
+import collections
+import os
+
+import pytest
+
+import moon_download
+from moon_download import FileRecord, RunFatalControl, download_file
+
+
+class FakeChunkIter:
+    """Yield the prepared chunks, recording every one it actually handed out.
+
+    `delivered` is an observation, not a model of the loop. A test that assumed
+    which chunks were consumed would keep passing if the order of the abort
+    check and the append ever changed, which is the thing worth catching.
+    """
+
+    def __init__(self, chunks, before_chunk=None, delivered=None, aborted=None, abort_probe=None):
+        self._chunks = list(chunks)
+        self._before_chunk = before_chunk
+        self.delivered = delivered if delivered is not None else []
+        self._aborted = aborted
+        self._abort_probe = abort_probe
+
+    def __aiter__(self):
+        return self
+
+    async def __anext__(self):
+        if self._before_chunk is not None:
+            self._before_chunk()
+        if not self._chunks:
+            raise StopAsyncIteration
+        chunk = self._chunks.pop(0)
+        self.delivered.append(chunk)
+        if self._aborted is not None:
+            self._aborted.append(bool(self._abort_probe()) if self._abort_probe else False)
+        return chunk
+
+
+class FakeContent:
+    def __init__(self, chunks, before_chunk=None, delivered=None, aborted=None, abort_probe=None):
+        self._chunks = chunks
+        self._before_chunk = before_chunk
+        self.delivered = delivered if delivered is not None else []
+        self._aborted = aborted
+        self._abort_probe = abort_probe
+
+    def iter_chunked(self, size):
+        return FakeChunkIter(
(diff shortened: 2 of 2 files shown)


LeyckerS (approved):
Verified rather than read, and every claim in the body held:

- **The diff is the one-line removal and the test file, nothing else** — `last_write_bytes`, `write_persisted_before` and the write itself untouched, exactly as described.
- **On the branch**: your 9 tests pass, and the full suite is 63/63.
- **On unmodified `main`**: your test file fails 6 and passes 3 — the exact split the body claims, with the doubling cases failing and the three deliberately-green pins (exact multiple, timestamp bound, abort observation) passing. That negative control is what makes the regression test worth its 289 lines.
- The test design deserves saying out loud: expectations read from the chunks the fake actually delivered rather than from a model of the loop, sizes derived from `WRITE_BUF`/`RECV_CHUNK` so the boundary cases survive a retune, and the exact-multiple case pinning the per-chunk append against a flush-site refactor. This is the shape I will point future test PRs at.
- On your `test_elapsed_clock` note: it passes here on both `main` and this branch (63/63), so it is your machine's timing, not your change — thank you for flagging it instead of hiding it.

Merging.

Issue #64, opened 2026-07-31
bug(gui): a square backdrop appears behind the progress rings during a run


LeyckerS:
### Context

Every transfer row shows a circular progress ring. While a run is in flight, a **square patch of a slightly different colour** appears behind each ring, roughly the size of the ring's box. It is visible on the actively downloading rows and it does not belong to the design.

Reported from real use in the WebView2 GUI. Screenshot in the report shows it on eight simultaneous `downloading` rows.

### Where to look

Two suspects, both in `web/styles.css`, and they interact.

**1. The drop-shadow filter on the ring (L598-606):**

```css
.ring .ring-fg {
  stroke: var(--c);
  ...
  filter: drop-shadow(0 0 4px color-mix(in srgb, var(--c) 80%, transparent));
  transition: stroke-dasharray .3s linear;
}
```

A CSS `filter` promotes the element to its own compositing layer. In Chromium-based engines that layer is rasterised against the element's bounding box, which is a rectangle — and a rectangle is exactly what is showing.

**2. `content-visibility` on the row (L564-577):**

```css
.frow {
  ...
  content-visibility: auto;
  contain-intrinsic-size: 3.4rem;
}
```

`content-visibility: auto` applies containment to the row. Containment plus a filtered descendant plus the `spin` animation on some states (L613-614) is a known source of repaint artifacts in Chromium.

The same `drop-shadow` pattern also exists on `.spark circle` (L514) — worth checking whether the sparkline shows a milder version of the same thing.

### What to do

**Bisect before fixing.** In DevTools, on a row that shows the artifact:

1. Disable `filter` on `.ring .ring-fg` — does the square disappear?
2. Re-enable it, disable `content-visibility` on `.frow` — does it disappear?
3. Report which one it was, or that it needs both.

Then fix the one that is responsible, keeping the glow if you can. Options worth trying, in order of preference:

- Move the filter to a wrapper that has no containment applied to it
- Replace `drop-shadow` with an SVG `<filter>` with an explicit oversized filter region
- Give the row `contain-intrinsic-size` room so the filter's bounds are not clipped
- Last resort: drop the glow on the ring. It is decoration; a square artifact is worse than no glow

**Do not** remove `content-visibility` outright without measuring — it is there for scroll performance with long transfer lists, and trading a visual glitch for a janky list is not an improvement. If you conclude it has to go, say so in the PR with what you observed.

### Acceptance criteria

[...]

PR #177, opened 2026-08-25
ci: add syntax check for web/ changes (fixes #174)

### LINKED ISSUE #174: PRs that touch only web/ run no CI at all — not even a syntax check
Both workflows filter by path, and `web/` matches neither:

- `.github/workflows/lint.yml` triggers on `**.py`, `requirements.txt`, `constraints.txt`, `ruff.toml`, `pytest.ini`, `.github/dependabot.yml`, and itself (`lint.yml:3-22`).
- `.github/workflows/docs-cli-check.yml` triggers on `**.md`, `**.py`, and itself.

So a pull request touching only `web/app.js`, `web/index.html` or `web/styles.css` arrives with an **empty check list**: no byte-compile, no lint, nothing. A missing brace in `app.js` would merge green — there is no green to withhold. #173, the current first-timer issue, even has to warn its taker that "no checks" is normal.

The gap is asymmetric with the rest of the repository: #164/#165 established the principle that every file controlling behaviour triggers CI, and the GUI — a third of what users touch — is the one part left out entirely.

## The minimal gate, verified

GitHub's `ubuntu-latest` runners ship Node.js, and `node --check web/app.js` parses the file without executing it — verified against current `main` (passes, Node 24). A one-job workflow with `paths: ["web/**"]` running that single command catches syntax errors — exactly the class of breakage a browser-only contributor is most likely to ship, since they cannot run the GUI's own checks locally.

## Worth discussing, not required

`render_gui.py` already exists as a real GUI check — headless-Chromium screenshots that fail on JS errors, element overflow, horizontal scroll, or missing rows — so wiring it into the same workflow (`pip install -r requirements.txt` + `playwright install chromium`) would gate far more than syntax. It is heavier (~1–2 minutes) and I have **not** verified it inside a runner, so treat it as a stretch goal: the syntax gate alone closes the "zero checks" hole and can land first, in its own small PR.

Size: one small workflow file. No Windows needed. To claim it, comment here.

AdriDob:
## Problem
PRs that touch only `web/**` run no CI at all — not even a syntax check (issue #174).

## Solution
Added a GitHub Actions workflow that runs `node --check web/app.js` on every push/PR affecting files under `web/**`.

## Testing
- Validated locally: `node --check web/app.js` passes
- Workflow triggers on `web/**` paths only (minimal scope)
- Uses Node.js 22 (current LTS) via `actions/setup-node@v4`

## Changes
- New file: `.github/workflows/web-syntax.yml`

### DIFF
--- a/.github/workflows/web-syntax.yml
+++ b/.github/workflows/web-syntax.yml
@@ -1,0 +1,27 @@
+name: Web Syntax Check
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "web/**"
+      - ".github/workflows/web-syntax.yml"
+  pull_request:
+    paths:
+      - "web/**"
+      - ".github/workflows/web-syntax.yml"
+
+jobs:
+  syntax-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+
+      - name: Syntax check web/app.js
+        run: node --check web/app.js


LeyckerS:
Thanks @AdriDob — and to be clear about why this is closing, because it is not the code: your workflow is correct, and your own CI run proves the trigger and the command both work. The reason is claim order. @tunglambk claimed #174 at 08:28 UTC on 25 August and opened #176 at 08:34; the claim on the issue and the open PR were both on the thread when you claimed at 17:15 and opened this four minutes later. Under the five-day claim rule this repo publishes (#39), #174 was already taken, and #176 has now been merged.

Some of this is my fault: the roadmap's free table still listed #174 as available — it was four days stale — and if you found the issue through the roadmap, you were set up for exactly this collision. Sorry about that; it is fixed now.

Two small technical notes, in case they are useful elsewhere: the setup-node step was unnecessary (ubuntu-latest ships Node 24, which is the current active LTS — 22 is in maintenance now), and the job id 'syntax-check' collides with the job of the same name in lint.yml, which would make a mixed .py+web PR's check list ambiguous.

If you want work that is genuinely free right now: #64 (a GUI backdrop bug) and #84 (a MiB/MB labelling mixup, just returned to the pool today) are both open. Comment on the issue first and it is yours — I answer every claim.

